PR #3, opened 2017-05-19
Fixes: ENOENT: no such file or directory, open '\node_vlocity\defaultdatapack.json'

thvd:
Fixes error, because the path `./node_vlocity/defaultdatapack.json` isn't always correct. Should be relative to the `__dirname` global module variable, that's why I have changed to use the `this.defaultDataPack` because that's read from the right relative path.


rutlabaga:
Thanks!

PR #5, opened 2017-07-03
Do not depend on node-sass and support NPM scripts for usage on windows

Codeneos:
These changes remove the dependency on node-sass which is hard to build on windows due to the dependency on node-gyp and python. Instead it uses sass.js which if native javascript version of the sass library to compile scss/sass into CSS.

This pull request also include an async compile queue extension that can be used to compile (postprocess) source code before adding it to the deployment pack. It is wirrten in such a way that it can be extended with to support other abstraction langues with little to no effort; for example [pug ](https://pugjs.org) or [haml](http://haml.info/).

Another part of this pull request focuses on making the deployment tools more easy to integrate into an existing build environment by adding support for NPM scripts. With this change set installation and running of build tools is reduced to just 1 line of code (assuming that you have node.js installed). Configure your properties and export file and run the deployment by just typing `npm run packDeploy -- -verbose -job data-pack-job -propertyfile <file>` from the console/terminal window.


Codeneos:
Great! In case of questions or comments let me know :)

rutlabaga:
Thanks so much for contributing!

Right now, there is an issue for me with "@import variables" in the Sass for a VlocityUITemplate when compiling Sass. I was able to solve this with node-sass by adding the option "includePaths" and passing in all of the folders in the directory.  I don't see a similar option in the sass.js documentation. Is there any way to compile in the @import's with this library?

Let me know if you need a repro.

Codeneos:
Yes there is an option to do that, but as I do not have any templates to test it with that actually use imports I was not able to test it. Do you have an example with imports for me to test it with? Then I can add the require changes and push an updated version to my branch :)

Codeneos:
Ok, I added support for imports and did a small test. I think it should be working now. Can @rutlabaga can you check if it is working for you as well?

Also I did some tuning in the way errors are reported back; I noticed that even when stuff went wrong grunt still reported 0 errors and finished saying the job completed successfully. Instead I changed the error handling to use grunt.warn so that you can see that an error occurred even when running a single job.

And I added a custom logo and title to the toast messages :).

rutlabaga:
Thanks so much again for this contribution. Very much appreciated! 

Definitely need to figure out how to add real tests to the project.

PR #7, opened 2017-08-03
Changed Grunt-Notify to Node-Notifier

Codeneos:
Grunt-Notify is not maintained anymore and caused not compatibility issue with the latest version of NPM. When using the Vlocity deployment tools on our CI/CD server we noticed that it crashed due to the grunt notify package not being able to load the default toast image.

To resolve the issue I replaced it with the Node-Notifier package which is better maintained and also flexible in the kind of notifications to show. This also allowed me to use a the custom toaster image a more robust way; no need for funky copies during start-up anymore.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,8 +1,6 @@
 
 'use strict';
 var loadProperties = require('./grunt/loadProperties.js');
-var path = require('path');
-var fs = require('fs');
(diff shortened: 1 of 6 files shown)


Codeneos:
@rutlabaga; were you able to review the pull request?

rutlabaga:
Thanks for the contribution!

PR #8, opened 2017-09-02
Improvements for APEX code runner and fix for SCSS compilation randomness

Codeneos:
I made some small improvements in the way APEX is loaded and executed:
- Better error handling when APEX code fails; error message will be shown and the build will be aborted.
- Properly wait for the APEX to execute before running the rest of a deployment job
  - there was an issue with the way the promises where used in the APEX code execution part which cause the promise rejection/resolve not to be returned and instead the APEX execution result object was returned.
- Better handling of errors; when an error occurs it is now logged a bit better and the notification is also shown.
- Fixed bug in SCSS compilation; when compiling SCSS into CSS sass.js would invoke the last passed callback for every compilation result instead of calling the passed callback. This is a bug in sass.js due to which compiling code in Parallel is not really possible. 
  - To resolve this problem SCSS is now compiled in sequence one after another this ensures that the right callbacks are invoked and the compile CSS is assigned to the correct template

### DIFF
--- a/grunt/dataPacksJobTask.js
+++ b/grunt/dataPacksJobTask.js
@@ -77,8 +77,6 @@ module.exports = function (grunt) {
 
 	    	vlocity.datapacksjob.runJob(dataPacksJobsData, jobName, action,
 	    		function(result) {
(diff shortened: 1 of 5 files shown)


rutlabaga:
Thanks again!

PR #28, opened 2018-01-16
Support proxy server configuration

Codeneos:
This add support for specifying a proxy server in the properties files. Readme was also updated to describe the new option. It is set using 'sf.httpProxy' from the properties files which is passed to the sf connect node library that will then proxy all request to salesforce through the proxy server.

We needed this as some of our clients run their CI/CD from servers that can only access the internet through a proxy.


rutlabaga:
Thanks!

PR #66, opened 2018-06-07
Add support for loading multiple pre-job and post-job apex files.

Codeneos:
This changes adds for support loading multiple APEX files as pre- and post-deployment job by specifying them in array syntax in the job yaml file:
```yaml
preJobApex:
  Deploy: 
   - apex/RemoveHandsetBundleCategory.cls
   - apex/RemoveVlocityActions.cls
  
postJobApex:
  Deploy: apex/PostDeployJob.cls
```

Also this pull request changes the way that the tools search for job files, using a array of possible folders to search for job files. This makes it easier to add new search directories in the future.

### DIFF
--- a/lib/datapacksbuilder.js
+++ b/lib/datapacksbuilder.js
@@ -69,11 +69,11 @@ DataPacksBuilder.prototype.buildImport = function(importPath, jobInfo, typesInAl
         nextImport = self.getNextImport(importPath, Object.keys(jobInfo.currentStatus), jobInfo.singleFile === true, jobInfo, typesInAllImports, currentDataPackKeysInImport,currentTypesInImport, !jobInfo.singleFile ? dataPackImportLength : 0);
 
         if (nextImport) {
-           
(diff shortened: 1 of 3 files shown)


rutlabaga:
It seems like this change is not actually replacing the BaseUtilities.cls during the preStepApex. If you do `runTestJob` then it is currently failing with:

 APEX Compilation Error >> Variable does not exist: dataPackDataSet

Because the:
//include BaseUtilities.cls;

Is still in the apexFileData when it sends to Salesforce.

Codeneos:
Oh, didnt notice that all during my testing. Will look at it today and push
fix.

Codeneos:
Update :) - can you have a look if it works?

I also added some chunking for executing anon apex. This significantly improved deployment performance for us as the main bottleneck for doing larger product catalog deployments was for us the pre- job APEX context data.

rutlabaga:
Thanks so much again!

PR #119, opened 2018-11-23
Feature/diffs api

rutlabaga:


lib/datapacksjob.js
```diff
@@ -560,18 +582,22 @@ DataPacksJob.prototype.runQueryForManifest = async function(queryInput) {
             this.vlocity.datapacksutils.updateRecordReferences(record);
 
             record.VlocityDataPackKeyForManifest = queryData.VlocityDataPackType + '/' + this.vlocity.datapacksexpand.getDataPackFolder(queryData.VlocityDataPackType, record.VlocityRecordSObjectType, record);
+            
+            if (jobInfo.fullManifest[queryData.VlocityDataPackType][record.Id] || 
+                (jobInfo.specificManifestKeys 
+                && jobInfo.specificManifestKeys.indexOf(queryData.VlocityDataPackType) == -1
```
thvd: instead of `indexOf` you can use the `includes` method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). Depends on which NodeJS version you want to support, if you can opt for this.

thvd: Based on the package.json file i see you support Node 8 or later. So `includes` is a valid option then.

PR #155, opened 2019-01-30
Activate embedded OS when reusable OS being deployed

kukkainna:
Activate embedded OS when reusable OS being deployed


lib/datapacksjob.js
```diff
@@ -1546,13 +1546,26 @@ DataPacksJob.prototype.activateAll = async function(dataPack, jobInfo, attempts)
 
             // If it is the only one in the deploy and it fails to activate it must be set to error. Otherwise retry the deploy and activation separate from others.
             if (dataPackData.dataPacks.length == 1) {
+                var onActivateErrorResult = {};
                 
+                onActivateErrorResult = await this.vlocity.datapacksutils.handleDataPackEvent('onActivateError', dataPack.VlocityDataPackType, dataPackData);
```
rutlabaga: This should just be 1 line

PR #161, opened 2019-02-05
Reactivate omniscripts when vlocity UI template is deployed 

kukkainna:
Reactivate omniscripts when vlocity UI template is deployed


lib/datapacktypes/vlocityuitemplate.js
```diff
@@ -0,0 +1,26 @@
+var async = require('async');
+
+var VlocityUITemplate = module.exports = function(vlocity) {
+	this.vlocity = vlocity;
+};
+
+VlocityUITemplate.prototype.onDeployFinish = async function(jobInfo) {
+    if (!this.vlocity.sfdxUsername) {
+        VlocityUtils.error('Deploy Error', 'Please run again using SFDX command');
```
rutlabaga: Can you make this message more clear? It should not be an error. Use VlocityUtils.report

VlocityUtils.report('Invalid Action', 'OmniScripts cannot be automatically activated unless using SFDX Authentication');

lib/datapacktypes/vlocityuitemplate.js
```diff
@@ -0,0 +1,26 @@
+var async = require('async');
+
+var VlocityUITemplate = module.exports = function(vlocity) {
+	this.vlocity = vlocity;
+};
+
+VlocityUITemplate.prototype.onDeployFinish = async function(jobInfo) {
+    if (!this.vlocity.sfdxUsername) {
```
rutlabaga: And can you check to make sure this isn't running for every VlocityUITemplate deployment, but for the ones related to OmniScripts based on the Type of the VlocityUITemplate.

rutlabaga: And it should be controlled by a setting to turn off.

lib/datapacktypes/vlocityuitemplate.js
```diff
@@ -0,0 +1,39 @@
...
+VlocityUITemplate.prototype.activateOmniScript = async function(inputMap) {
+    var omniscript = inputMap.omniscript;
+    await this.vlocity.datapacksutils.jsRemote(`${this.vlocity.namespace}.EmbeddingBusinessProcessController`, 'activateScript', [ omniscript.Id, 1], `${this.vlocity.namespace}__embeddingbusinessprocesspage`)
+}
+
+VlocityUITemplate.prototype.onActivate  = async function(inputMap) {
+    var jobInfo = inputMap.jobInfo;
+    var dataPack = inputMap.dataPack;
+
+    if (jobInfo.reActivateOmniScripts || typeof(jobInfo.reActivateOmniScripts) === 'undefined') {
```
rutlabaga: should only happen when the value is true, not when undefined.

PR #166, opened 2019-02-12
Export dependency for VlocityUITemplates automatically

kukkainna:
Get All SCSS Files as a Dependency for VlocityUITemplates Automatically


lib/datapacksexpand.js
```diff
@@ -704,6 +704,7 @@ DataPacksExpand.prototype.updateSourceKeysInDataPack = function(currentData, job
 DataPacksExpand.prototype.processDataPack = function(dataPackData, jobInfo, isPagination) {
 
     var self = this;
+    self.jobInfo = jobInfo;
```
rutlabaga: Don't add this. Just pass to next method.

lib/datapacksexpand.js
```diff
@@ -962,6 +964,38 @@ DataPacksExpand.prototype.processDataPackData = function(dataPackType, parentNam
     }
 };
 
+DataPacksExpand.prototype.addDataPackToFullManifest = function(extension, dataPackType, expansionData) {
```
rutlabaga: This method should be split into components of what it does. Get referenced datapacks by scss. Then an add to jobfile method should be in datapacksjob.js or datapacksutils.js.

lib/datapacksexpand.js
```diff
@@ -962,6 +964,38 @@ DataPacksExpand.prototype.processDataPackData = function(dataPackType, parentNam
...
+    var self = this;
+
+    if (extension === 'scss'
+    && self.jobInfo.compileOnBuild 
+    && self.jobInfo.maxDepth != 0) {
+        var regex = new RegExp("(?:@import \")(.*?)(?:;)", "ig");
+        var found = expansionData.match(regex);
+        var templates = [];
+
+        for (var key in self.jobInfo.allDataSummary) {
```
rutlabaga: This isn't really needed. You could just get the templates from the regex then add them to fullManifest if fullManifest doesn't already contain the VlocityDataPackKey

lib/datapacksexpand.js
```diff
@@ -962,6 +963,39 @@ DataPacksExpand.prototype.processDataPackData = function(dataPackType, parentNam
...
+        if (!jobInfo.compileOnBuild) {
+            return;
+        }
+
+        var regex = new RegExp("(?:@import \")(.*?)(?:;)", "ig");
+        var found = expansionData.match(regex);
+
+        if (found) {
+
+            if (!jobInfo.fullManifest) {
```
rutlabaga: If this is null then just return since it is not an export anyway.

PR #183, opened 2019-03-04
Auto deploy picklist values on deploy fixes global and not

kukkainna:
Auto deploy picklist values on deploy fixes global and not


lib/deltacheck.js
```diff
@@ -99,70 +103,77 @@ DeltaCheck.prototype.executeAutoFixPicklists = async function() {
...
 
-                        if (!valueFound) {
-                            fieldValues.push({fullName: dataPackPickListValue, default: false, isActive: true, label: dataPackPickListValue});
+                            valueNotFound = false;
                         }
                     }
+
+                    if (valueNotFound) {
+                        updateCustomObject = true;
+                        fieldValues.push({fullName: dataPackPicklistValues[y], label: dataPackPicklistValues[y], default: false, isActive: true});
```
rutlabaga: Will this add the same value more than once if it is in multiple datapacks?

PR #184, opened 2019-03-05
Alex/suppress display sequence

kukkainna:
Support display sequence auto fix on attribute category object


rutlabaga:
Switch to dx branch

PR #186, opened 2019-03-08
Find duplicate matching keys and update it, if a global key is null add it

kukkainna:
Find duplicate matching keys and update it, if a global key is null add it

### DIFF
--- a/javascript/cleanData.js
+++ b/javascript/cleanData.js
@@ -1,8 +1,6 @@
 var async = require('async');
 
-module.exports = function(vlocity, currentContextData, jobInfo, callback) {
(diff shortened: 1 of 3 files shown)


javascript/cleanData.js
```diff
@@ -1,8 +1,6 @@
 var async = require('async');
 
-module.exports = function(vlocity, currentContextData, jobInfo, callback) {
-
-    // Add Global Key
+module.exports = async function(vlocity, currentContextData, jobInfo, callback) {
```
rutlabaga: Probably should not make this async

PR #187, opened 2019-03-08
Alex/suppress display sequence

kukkainna:
Auto fix display sequence errors on data pack deploy

### DIFF
--- a/lib/datapacksbuilder.js
+++ b/lib/datapacksbuilder.js
@@ -81,7 +81,7 @@ DataPacksBuilder.prototype.buildImport = async function(importPath, jobInfo) {
     var shouldBreakImportLoop = false;
 
     do {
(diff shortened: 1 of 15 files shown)


lib/datapacksbuilder.js
```diff
@@ -326,7 +326,20 @@ DataPacksBuilder.prototype.setFileData = async function(filePath, filePath, enco
...
+
+    try {
+        var dataPack = JSON.parse(data);
+
+        if (dataPack && dataPack.VlocityRecordSourceKey) {
+            if (!this.recordSourceKeyToFilePath) {
+                this.recordSourceKeyToFilePath = {};
+            }
+
+            this.recordSourceKeyToFilePath[dataPack.VlocityRecordSourceKey] = path.normalize(filePath);
```
rutlabaga: There is more than 1 file per dataPack.VlocityRecordSourceKey, so this will not be a correct flow.

PR #207, opened 2019-06-19
backend  changes for vpl support in vbt

vswaroop:
Added changes to support VPL in build tool UI

### DIFF
--- a/lib/datapacksjob.js
+++ b/lib/datapacksjob.js
@@ -2599,75 +2599,65 @@ DataPacksJob.prototype.deploySalesforce = async function(jobInfo) {
     await this.vlocity.datapacksutils.deploySalesforce(jobInfo);
 }
 
-DataPacksJob.prototype.rmdirCustom = async function(dir) {
-	var list = fs.readdirSync(dir);
-	for(var i = 0; i < list.length; i++) {
-		var filename = path.join(dir, list[i]);
-		var stat = fs.statSync(filename);
+DataPacksJob.prototype.rmdirCustom = async function (dir) {
+    var list = fs.readdirSync(dir);
+    for (var i = 0; i < list.length; i++) {
+        var filename = path.join(dir, list[i]);
+        var stat = fs.statSync(filename);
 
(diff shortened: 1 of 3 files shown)


lib/datapacksjob.js
```diff
@@ -2591,3 +2598,130 @@ DataPacksJob.prototype.retrieveSalesforce = async function(jobInfo) {
...
+            VlocityUtils.verbose("fetched-2 : " + result.records.length);
+            VlocityUtils.verbose(VplAssetRecords);
+            jobInfo.data = result;
+        });
+
+        for(res in VplAssetRecords)  {
+            let query = "SELECT Name, Body FROM Attachment WHERE ParentId = '" + VplAssetRecords[res].VPL_Asset__c + "'";
+            VlocityUtils.verbose('query:');
+            VlocityUtils.verbose(query);
+            await this.vlocity.jsForceConnection.query(query, function (err, result) {
```
rutlabaga: You should use the try / catch syntax for these instead of a callback. Better for readability.

lib/datapacksjob.js
```diff
@@ -2591,3 +2598,130 @@ DataPacksJob.prototype.retrieveSalesforce = async function(jobInfo) {
...
+            VplAssetRecords = result.records;
+            VlocityUtils.verbose("total-2 : " + result.totalSize);
+            VlocityUtils.verbose("fetched-2 : " + result.records.length);
+            VlocityUtils.verbose(VplAssetRecords);
+            jobInfo.data = result;
+        });
+
+        for(res in VplAssetRecords)  {
+            let query = "SELECT Name, Body FROM Attachment WHERE ParentId = '" + VplAssetRecords[res].VPL_Asset__c + "'";
+            VlocityUtils.verbose('query:');
```
rutlabaga: `VlocityUtils.verbose('query:', query);`

Keep all log statements to 1 line.

lib/datapacksjob.js
vswaroop: ok

PR #214, opened 2019-07-22
Add support for git branch command to list all available branches

vswaroop:


lib/datapacksjob.js
```diff
@@ -441,16 +443,10 @@ DataPacksJob.prototype.formatResponse = async function(jobInfo, error) {
             response.status = jobInfo.hasError ? 'error' : 'success';
             response.message = jobInfo.hasError ? jobInfo.errors.join('\n').replace(/%vlocity_namespace%/g, this.vlocity.namespace) : `Retrieve Successful`;
             response.records = jobInfo.sfdxData;
-        } else if (jobInfo.jobAction == 'GitCommit') {
+        } else if (jobInfo.jobAction.startsWith('Git') || jobInfo.jobAction == 'refreshVlocityProcessListing') {
```
rutlabaga: I don't think these should be combined. Since the message is not the same for both.

vswaroop: Ok. actually I was more concerned about the failure message which would be different in each case. I am handling the success messages at the frontend.

PR #218, opened 2019-08-01
handle commit action inside migrate api

vswaroop:
Also fixed namespace issue due to which vpl download was failing.


lib/utilityservice.js
```diff
@@ -359,6 +359,7 @@ UtilityService.prototype.getNamespace = async function() {
         } catch (err) {
             if (this.vlocity.passedInNamespace) {
                 VlocityUtils.verbose('Expected Namespace Failure', err, this.vlocity.passedInNamespace );
+                return;
```
rutlabaga: This `return` should be removed. The code below this still needs to run.

PR #219, opened 2019-08-07
fix commit with dependencies

vswaroop:


lib/utilityservice.js
```diff
@@ -831,16 +830,16 @@ UtilityService.prototype.runGitCommit = async function (jobInfo) {
...
     }
     await self.runGitInit(jobInfo);
 
-    if (Array.isArray(jobInfo.fileName)) {
-        for (var fileOrFolderName in jobInfo.fileName) {
+    if (Array.isArray(jobInfo.manifest)) {
+        for (var fileOrFolderName in jobInfo.manifest) {
 
-            var sanitizedFileorFoldername = unidecode(jobInfo.fileName[fileOrFolderName])
+            var sanitizedFileorFoldername = unidecode(jobInfo.manifest[fileOrFolderName])
```
rutlabaga: This fails now when the VlocityDataPackKey points to Salesforce Metadata because these are not sanitized the same way. Probably best to first check that the file / folder exists unsanitized then check to see if folder exists sanitized.

vswaroop: Ok.

PR #223, opened 2019-09-03
Add git status command and git pull fix for remote branches

vswaroop:


lib/utilityservice.js
```diff
@@ -1211,3 +1246,8 @@ UtilityService.prototype.updateSObjectsBulk = async function(sObjectType, update
 
     await this.parallelLimit(queryPromises, 5);
 };
+
+UtilityService.prototype.ignoreTerminalPrompt = async function (dirname) {
+    childProcess.execSync(`cd ${dirname}`);
```
rutlabaga: What does cd dirname do in this case? It seems like setting the environment variable is happening in the process. Is this cd necessary?

vswaroop: Apparently this env variable is getting set for the terminal prompt in current directory. If I don't change the dirname, it ends up setting it in the repository folder terminal prompt only and we don't get the desired result.

PR #235, opened 2019-10-29
BTCA-215  Auto Export Apex Classes from Vlocity DX

ManasAgarwalVlocity:


dataPacksJobs/AllDefault.yaml
```diff
@@ -60,4 +60,5 @@ defaultMaxParallel: 10
 exportPacksMaxSize: 10
 useAllRelationships: false
 supportHeadersOnly: true
-supportForceDeploy: true
\ No newline at end of file
+supportForceDeploy: true
+includeSalesforceMetadata: true
```
rutlabaga: Don't include this

lib/datapacksexpand.js
```diff
@@ -425,6 +425,17 @@ DataPacksExpand.prototype.getNestedDataPackReferences = function(currentData, pa
                 }
                 
             } else {
+                if(!(jobInfo.maxDepth === 0) || jobInfo.includeSalesforceMetadata){
+                    let referenceInfo =  this.vlocity.datapacksutils.handleDataPackEvent('onAutoExport', referenceType, currentData);
+                    if (!jobInfo.specificManifestKeys){
+                        jobInfo.specificManifestKeys = [];
+                    } 
+                    referenceInfo.then(result => {
```
rutlabaga: better to use `await` on line 429 to make it more clear

dataPacksJobs/AllDefault.yaml
ManasAgarwalVlocity: Removed "includeSalesforceMetadata"

lib/datapacksexpand.js
ManasAgarwalVlocity: Changed to await.

lib/datapacksexpand.js
```diff
@@ -438,10 +447,10 @@ DataPacksExpand.prototype.getNestedDataPackReferences = function(currentData, pa
 
         if (Array.isArray(currentData[primaryField])) {
             for (var nestedMore of currentData[primaryField]) {
-                this.getNestedDataPackReferences(nestedMore, parentKeys, splitFields.join('.'), referenceType);
+                this.getNestedDataPackReferences(jobInfo, nestedMore, parentKeys, splitFields.join('.'), referenceType);
```
rutlabaga: Since `async` was added to this method, you'll need to add `await` to these `getNestedDataPackReferences` calls.

lib/datapacksexpand.js
```diff
@@ -425,6 +425,15 @@ DataPacksExpand.prototype.getNestedDataPackReferences = function(currentData, pa
                 }
                 
             } else {
+                if(!(jobInfo.maxDepth === 0) || jobInfo.includeSalesforceMetadata){
```
rutlabaga: This should probably be an && where `(jobInfo.maxDepth != 0 && jobInfo.includeSalesforceMetadata)` since if either false then it should not run.

rutlabaga: This should actually just be `if (jobInfo.maxDepth != 0)` - Then in the apexclass.js it should check the 

`jobInfo.includeSalesforceMetadata` to make sure of whether or not to return a valid value. If jobInfo.includeSalesforceMetadata is false it could just return null.

lib/datapacksexpand.js
ManasAgarwalVlocity: Updated

lib/datapacksexpand.js
ManasAgarwalVlocity: Code has been updated but Why are we waiting for apexclass.js to return null, if it can be skipped in datapacksexpand.js itself?

rutlabaga: Because the check for `jobInfo.includeSalesforceMetadata` only applies to Salesforce Metadata not Vlocity DataPacks. So if a Vlocity DataPack onAutoExport handler were added then this check would no longer be valid there.

lib/datapacktypes/apexclass.js
```diff
@@ -0,0 +1,11 @@
+var ApexClass = module.exports = function(vlocity) {
+    this.vlocity = vlocity;
+};
+
+ApexClass.prototype.onAutoExport = function(refObject) {
+    if(refObject.jobInfo.includeSalesforceMetadata && !refObject.currentData.remoteClass.includes("%vlocity_namespace%")){
+        return 'classes/'+ refObject.currentData.remoteClass + '.cls';
+    } else{
```
rutlabaga: This is not actually necessary. If you have no return it will return null automatically.

PR #241, opened 2019-11-26
BTCA-325 Add ability to open a Vlocity Component in its own Designer 

ManasAgarwalVlocity:


lib/datapacktypes/omniscript.js
```diff
@@ -10,24 +10,27 @@ var OMNI_BULK_RECORDS_LIMIT_MIN = 1000;
...
+    let dataField = '%vlocity_namespace%__OmniScript__c';
+    if(!dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c']){
+        dataField = 'OmniScript__c';
+    }
+    if (dataPackData.VlocityDataPackData[dataField][0]['%vlocity_namespace%__Element__c']) {
+        dataPackData.VlocityDataPackData[dataField][0]['%vlocity_namespace%__Element__c'] = this.vlocity.datapacksexpand.sortList(dataPackData.VlocityDataPackData[dataField][0]['%vlocity_namespace%__Element__c'], '%vlocity_namespace%__OmniScript__c');
     }
     
-    var elements = dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c'];
+    var elements = dataPackData.VlocityDataPackData[dataField][0]['%vlocity_namespace%__Element__c'];
```
rutlabaga: Can you switch it to using a variable for %vlocity_namespace%__Element__c and all the field names so that it is consistent similar to dataField where if dataField needs a namespace then they will all need a namespace, but if not then they all won't.

lib/datapacksutils.js
```diff
@@ -1454,7 +1462,11 @@ DataPacksUtils.prototype.getTitleFieldsForDataPack = async function(sourceDataPa
     var targetPrimarySObject;
     
     if (targetDataPack) {
-        targetPrimarySObject = targetDataPack.VlocityDataPackData[dataField][0];
```
rutlabaga: Would be better to just call this.getDataField(targetDataPack); again for targetDataPack

lib/datapacksutils.js
```diff
@@ -1305,7 +1309,11 @@ DataPacksUtils.prototype.getFieldDiffsIndividual = async function(allSourceDataP
                 var allTargetSObjects = [];
 
                 if (targetDataPack) {
-                    allTargetSObjects = this.getAllIndividualSObjectsInDataPack(targetDataPack.VlocityDataPackData[dataField][0]);
+                    if(targetDataPack.VlocityDataPackData[dataField]){
```
rutlabaga: Would be better to just call this.getDataField(targetDataPack); again for targetDataPack

lib/datapacksjob.js
```diff
@@ -538,7 +538,32 @@ DataPacksJob.prototype.formatResponse = async function(jobInfo, error) {
                         VlocityUtils.verbose(dataPack.VlocityDataPackKey, dataPack.VlocityDataPackStatus); 
                     }
                 }
-
+                if(Object.keys(allManifestByKey).length !== 0){
```
rutlabaga: This chunk should be in `formatResponse` where it can be a part of the other DataPack data getting injected in the large `else` statement. It should be as straightforward as if there is an "Id" then add the orgUrl. This spot is similar but then you are iterating over the manifest vs sourceData, but this should not be necessary as it already happens in formateResponse.

lib/utilityservice.js
```diff
@@ -1304,3 +1304,13 @@ UtilityService.prototype.ignoreTerminalPrompt = async function (dirname) {
     childProcess.execSync(`cd ${dirname}`);
     process.env.GIT_TERMINAL_PROMPT = 0;
 }
+
+UtilityService.prototype.createVisualForceOrgUrl = function (urlTemplate, datapackId) {
+    let designerPath = '';
```
rutlabaga: Should probably move all the logic for creating the Url here. It should just send the dataPackData here and then return the fully formed Url from this single method. Since otherwise there is duplicate code to get the urlTemplate and dataPackId.

lib/datapacktypes/omniscript.js
```diff
@@ -10,24 +10,27 @@ var OMNI_BULK_RECORDS_LIMIT_MIN = 1000;
 OmniScript.prototype.createTitleObjects = function(input) {
     var dataPackData = input.dataPackData;
     var titleObjects = [];
-
-    if (dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c']) {
-        dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c'] = this.vlocity.datapacksexpand.sortList(dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c'], '%vlocity_namespace%__OmniScript__c');
+    let dataField = '%vlocity_namespace%__OmniScript__c';
+    if(!dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c']){
+        dataField = 'OmniScript__c';
```
rutlabaga: This check is good, but all the other field names should have their namespace also removed right here if this one is.

lib/datapacktypes/omniscript.js
```diff
@@ -10,24 +10,37 @@ var OMNI_BULK_RECORDS_LIMIT_MIN = 1000;
 OmniScript.prototype.createTitleObjects = function(input) {
     var dataPackData = input.dataPackData;
     var titleObjects = [];
-
-    if (dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c']) {
-        dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c'] = this.vlocity.datapacksexpand.sortList(dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c'][0]['%vlocity_namespace%__Element__c'], '%vlocity_namespace%__OmniScript__c');
+    let omniScriptDataField = '%vlocity_namespace%__OmniScript__c',
+        element_c_DataField = '%vlocity_namespace%__Element__c';
+    if(!dataPackData.VlocityDataPackData['%vlocity_namespace%__OmniScript__c']){
+        omniScriptDataField = 'OmniScript__c';
```
rutlabaga: You can just do this one check then set all the fields to not use the %vlocity_namespace%. If 1 doesn't have namespace then none of them have it.

rutlabaga: Or you could just use the ${namespace} syntax in this file.

ManasAgarwalVlocity: I was considering that primary and secondary keys and their functionality may or may not be same in case of namespace.

lib/datapacksjob.js
ManasAgarwalVlocity: Iterations have been removed from formatResponse function.

lib/datapacksjob.js
```diff
@@ -1248,7 +1253,9 @@ DataPacksJob.prototype.exportGroup = async function(inputMap) {
                                 }
                             });
                         }
+                        dataPack.VlocityDataPackData[dataField][0].orgUrl =  this.vlocity.utilityservice.getVisualForcePageUrl(dataPack.VlocityDataPackType, dataPack.VlocityDataPackData[dataField][0].Id);
                     }
+                    jobInfo.VlocityDataPackKeyToUrlMapping[dataPack.VlocityDataPackKey] = this.vlocity.utilityservice.getVisualForcePageUrl(dataPack.VlocityDataPackType, dataPack.VlocityDataPackData[dataField][0].Id);
```
rutlabaga: Seems like this needs to be in the `if (dataField` also

lib/datapacksjob.js
```diff
@@ -486,6 +485,13 @@ DataPacksJob.prototype.formatResponse = async function(jobInfo, error) {
                     dataPack.Diffs = dataPack.DiffType != 'Unchanged';
                 }
 
+                if (!dataPack.orgUrl && jobInfo.VlocityDataPackKeyToUrlMapping[dataPack.VlocityDataPackKey]) {
+                    dataPack.orgUrl = jobInfo.VlocityDataPackKeyToUrlMapping[dataPack.VlocityDataPackKey];
+                    if(dataPack.orgUrl){
+                        dataPack.orgUrl = dataPack.orgUrl.replace(/%vlocity_namespace%__/g,this.vlocity.namespacePrefix);
```
rutlabaga: I don't think you need this replace because it is already in the `getVisualForcePageUrl` function.

ManasAgarwalVlocity: Though its getting replaced while debug, but in real time run, its coming as "%vlocity_namespace%"

rutlabaga: Ok, seems good then.

PR #245, opened 2019-12-06
fetch username from userId

vswaroop:


lib/datapacksjob.js
```diff
@@ -2769,3 +2775,23 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
     }
 }
+
+DataPacksJob.prototype.getUserNameFromUserId = async function (jobInfo) {
+    var query1;
+    jobInfo.data = [];
+    var records =  [];
+    try {
+        for(var i = 0;i<jobInfo.userId.length;i++)  {
+            query1  = `SELECT Id, Name FROM User WHERE Id = '${jobInfo.userId[i]}'`;
```
rutlabaga: Can you change this to a single Query. SELECT Id, Name FROM User Limit 10000. Then for any more than that it is fine to just show the Id in the UI.

vswaroop: Sure.

PR #247, opened 2019-12-10
Fix query definition for ContentVersion object

Codeneos:
`ContentVersion` objects do not have a `Name `field` and only have a title, removed the name field from  the query definition to fix query errors when exporting just ContentVersion objects.


rutlabaga:
Thanks!

PR #249, opened 2019-12-10
Fix case typo in UtilityService.getPackageVersion

Codeneos:
The code in `UtilityService.getPackageVersion` should only run if we do not yet know the PackageVersion, but it checks if `this.vlocity.packageVersion` is set which is never being set, instead `this.vlocity.PackageVersion` is being set (see line 411)

I expect the the intend from the author must have been not to run query the package version multiple times when the package version is already known.


rutlabaga:
Thanks!

PR #250, opened 2019-12-10
BTCA-246 Check Matching Keys Against Each Other For 2 Orgs

ManasAgarwalVlocity:


lib/datapacksjob.js
```diff
@@ -2150,6 +2151,19 @@ DataPacksJob.prototype.deployJob = async function(jobInfo) {
         jobInfo.resetFileData = true;
     }
 
+    if(Object.keys(jobInfo.sourceKeyToMatchingKeysData).length === 0){
+        var fullDataPath = jobInfo.projectPath;
```
rutlabaga: When you do this part you will need to clone the jobInfo to a new object otherwise these settings will stay with it. This is causing the issue in the build org. Do a `JSON.parse(JSON.stringify(jobInfo))` into a new Object and use that to send to await this.vlocity.datapacksbuilder.buildImport - Then when processing the result add it to the actual jobInfo object.

ManasAgarwalVlocity: Resolved

rutlabaga:
this has been merged.

PR #259, opened 2020-01-20
BTCA_335_Version Comparison feature

ManasAgarwalVlocity:


lib/datapacksjob.js
```diff
@@ -763,10 +763,15 @@ DataPacksJob.prototype.buildManifestFromQueries = async function(jobInfo) {
 
                 for (var query of jobInfo.queries) {
                     if (allTypesForQuery.has(query.VlocityDataPackType)) {
+                        if(query.VlocityDataPackType.includes('Version')){
```
rutlabaga: Should be removable, right?

lib/datapacksjob.js
```diff
@@ -780,8 +785,9 @@ DataPacksJob.prototype.buildManifestFromQueries = async function(jobInfo) {
                 if (!queryData || (!jobInfo.specificManifestKeys && queryData.manifestOnly)) {
                     continue;
                 }
-
-                allQueries.push( { context: this, func: 'runQueryForManifest', argument: { jobInfo: jobInfo, queryData: queryData } });
+                if((jobInfo.versionCompare && queryData.versionCompare) || (!jobInfo.versionCompare && !queryData.versionCompare)){
```
rutlabaga: Should just change this to return early `if (queryData.versionCompare && !jobInfo.versionCompare)`

lib/datapacksjob.js
```diff
@@ -763,10 +763,15 @@ DataPacksJob.prototype.buildManifestFromQueries = async function(jobInfo) {
...
                 for (var query of jobInfo.queries) {
                     if (allTypesForQuery.has(query.VlocityDataPackType)) {
+                        if(query.VlocityDataPackType.includes('Version')){
+                            query.VlocityDataPackType = query.VlocityDataPackType.replace('Version','');
+                        }
                         validQueries.push(query);
                     }
                 }
-
+                if(jobInfo.specificManifestKeys.length === 1 && jobInfo.specificManifestKeys[0].includes('Version')){
```
rutlabaga: Should also be removed?

lib/datapacksjob.js
ManasAgarwalVlocity: If its not removed, it will leave the trace everywhere in all the jsons exported,  against the field name "VlocityDataPackType"

lib/datapacksjob.js
ManasAgarwalVlocity: Updating the manifest keys so as to filter the datapacks correctly after fetching records.

PR #262, opened 2020-02-07
BTCA-360 VBT Analytics

ManasAgarwalVlocity:


lib/vlocity.js
```diff
@@ -125,8 +125,12 @@ Vlocity.runDataPacksCommand = async function(action, options) {
 
             process.chdir(options.chdir);
         }
+        var passedInOptions = JSON.parse(JSON.stringify(options)),
+            vlocity_run = new Vlocity(passedInOptions);
```
rutlabaga: Definitely this should not be moved like this.

rutlabaga: Definitely this new Vlocity should not be moved here.

rutlabaga: All the ones in this file are for VBT and can be ignored anyway.

ManasAgarwalVlocity: Changes have been made in vlocityCli.js to track actual command provided by user.  Now UI changes to send config variable(disableVbtAnalytics) can be ignored.

lib/vlocitycli.js
```diff
@@ -779,3 +785,12 @@ VlocityCLI.prototype.runTestJob = async function(dataPacksJobData) {
     
     return { message: message, status: 'success', action: 'Test Job' };
 }
+
+VlocityCLI.prototype.updateAnalytics = function (eventAction) {
```
rutlabaga: There still needs to be a way to disable the analytics from the jobInfo and a command line argument.

PR #269, opened 2020-03-05
BTCA-218 Where Used field in datapack diffs for reference

ManasAgarwalVlocity:


lib/datapacksutils.js
```diff
@@ -632,16 +632,29 @@ DataPacksUtils.prototype.getAllParentKeys = async function(dataPacks, jobInfo) {
...
 
-            if ((foundParentKey != dataPack.VlocityDataPackKey 
-                && jobInfo.currentStatus[foundParentKey] 
-                && newParents.indexOf(foundParentKey) == -1) 
+            if ((foundParentKey != dataPack.VlocityDataPackKey
+                && jobInfo.currentStatus[foundParentKey]
+                && newParents.indexOf(foundParentKey) == -1)
                 || (foundParentKey.startsWith('classes/'))) {
                 newParents.push(foundParentKey);
             }
```
rutlabaga: This is still N*N in the next section. You should be able to build the Child References for each DataPack by key in this loop. Each time a new Parent is `foundParentKey == true` that means the Child and parent both exist and the Parent has this dataPack as its child. So you could do something like `newChildReferences[foundParentKey].push(dataPack.VlocityDataPackKey)`. Then in the next loop you just need to iterate over the dataPacks again and set the `dataPack.VlocityDataPackChildren = newChildReferences[ dataPack.VlocityDataPackKey]`.

PR #274, opened 2020-03-26
BTCA-402 packDeployMultiple for several project paths

ManasAgarwalVlocity:


lib/datapacksjob.js
```diff
@@ -2418,6 +2420,20 @@ DataPacksJob.prototype.deployJob = async function(jobInfo) {
     VlocityUtils.milestone('Finalizing Deployment');
 };
 
+DataPacksJob.prototype.deployMultipleJob = async function (jobInfo) {
+    try {
+        for (const i in jobInfo.projectPaths) {
+            await fs.copy(jobInfo.projectPaths[i], jobInfo.projectPath);
```
rutlabaga: You will have to combine these in a separate folder as the original, since that will cause issues after deploy where they are now mixed without knowing which is original. You could use `vlocity-temp/combined` as the combined folder. Also, it will have to take into account different Expansion paths as well. For this command specifically you might want to use a new key in the job file.

PR #277, opened 2020-04-13
BTCA-290 Salesforce Metadata Dependencies auto-retrieval

ManasAgarwalVlocity:


lib/datapacksutils.js
```diff
@@ -3042,8 +3042,90 @@ DataPacksUtils.prototype.loadSalesforceMetadata = async function() {
 
 var folderOnlyMetadata = [ "aura", "layouts", "lwc", "staticresources" ]
 
-DataPacksUtils.prototype.retrieveSalesforce = async function(jobInfo, listOnly) {
+DataPacksUtils.prototype.getAdditionalSFMetadataDependencies = async function(jobInfo){
+    let vlocityDataPackTypes = [],
+        salesforceKeys = [],
+        salesforceTypes = [];
+
+    Object.keys(jobInfo.queries).forEach(key => {
```
rutlabaga, line 3050: This is not really an accurate way to get the Vlocity vs Salesforce types. Would probably be better to write a utility method that gets all the actual supported types from the Org in the VlocityDataPackConfiguration__mdt object depending on why this needs to happen.

lib/datapacksutils.js
```diff
@@ -3998,6 +4173,11 @@ DataPacksUtils.prototype.getSfdxData = async function(jobInfo) {
         checkAllowedTypes = true;
     }
 
+    let sfdxReferences = {};
+    if (fs.existsSync(path.join(sfdxProject.sfdxPath, 'sfReferences.yaml'))) {
```
rutlabaga, line 4177: This should just be spelled out as SalesforceReferences.yaml

lib/utilityservice.js
```diff
@@ -1359,4 +1359,18 @@ UtilityService.prototype.getVisualForcePageUrl = function (dataPackType, datapac
...
     return designerPath;
+}
+
+UtilityService.prototype.fetchAllRecords =  async function(query){
+    try {
+        var fetchedRecords = [];
+        var result = await this.vlocity.jsForceConnection.tooling.query(query);
+        fetchedRecords = fetchedRecords.concat(result.records);
+        if (result.records.length === 2000) {
+            fetchedRecords = fetchedRecords.concat(await this.fetchAllRecords(query + ' OFFSET 2000'));
```
rutlabaga, line 1370: Did you test that offset works here?

lib/datapacksutils.js
```diff
@@ -3042,8 +3042,115 @@ DataPacksUtils.prototype.loadSalesforceMetadata = async function() {
...
+        return vlocityDataPackTypes.indexOf(item.split('/')[0]) === -1;
+    });
+
+    salesforceTypes = Array.from(new Set(
+        salesforceKeys.map(item => {
+            return item.split('/')[0];
+        })
+    ));
+
+    if (salesforceTypes.indexOf('classes') !== 0) {
```
rutlabaga, line 3062: It would be good to make this more generic so it can work for other types beside ApexClass if we want to add them. This whole part with Apex classes here and when doing the MetadataComponentDependency check can probably be made into a function that accepts type inputs for everything.

lib/datapacksutils.js
```diff
@@ -3042,8 +3042,115 @@ DataPacksUtils.prototype.loadSalesforceMetadata = async function() {
...
+        var allApexClassData = await this.vlocity.utilityservice.fetchAllRecords('SELECT Id,Name FROM ApexClass');
```
rutlabaga, line 3063: Here you can change it to where NamespacePrefix = null - fine for this feature to only work in managed package.

lib/datapacksutils.js
```diff
@@ -3584,6 +3704,61 @@ DataPacksUtils.prototype.parallelRetrieve = async function(input) {
...
+
+    let deployments = {
+        targetusername: input.deployments.targetusername
+    };
+
+    var retrievedLWCDependencyKeys = [],
+        inboundFiles = [];
+
+    htmlFileContent.forEach(item => {
+        htmlContentArr = item.split('<');
```
rutlabaga, line 3728: should use let or var

lib/utilityservice.js
ManasAgarwalVlocity: Yes. It worked here as in my org it has 2568 ApexClasses, So it fetched in sets of 2000 & 568

lib/datapacksutils.js
```diff
@@ -3042,8 +3042,135 @@ DataPacksUtils.prototype.loadSalesforceMetadata = async function() {
...
+        }
+    });
+
+    return newSFKeysToBeAddedToManifest;
+}
+
+DataPacksUtils.prototype.retrieveExceptionalMetadataDependencies = async function(jobInfo, salesforceTypes) {
+    var exceptionDetailsObj = {},
+        dependenciesArr = [],
+        exceptionsArr = ['ApexClass'];
```
rutlabaga, line 3135: You can just hardcode the instantiation of the exceptionsArr. The rest of the function is perfect for adding a second type without having to re-code it all. But this first part can still be "hardcoded"

PR #279, opened 2020-04-23
BTCA-412 Profile Deployment Fix

ManasAgarwalVlocity:


lib/datapacksutils.js
```diff
@@ -4146,31 +4147,47 @@ DataPacksUtils.prototype.deploySalesforce = async function(jobInfo) {
     }
 
     if (deploymentOptions.sourcepath) {
-        
-        await this.prepareProfilesForDeploy(deploymentOptions.sourcepath);
+        let leadingPath = path.join('main', 'default'),
+            tempSfdxFolder = path.join(jobInfo.tempFolder, 'tempDeploy', 'salesforce_sfdx');
```
rutlabaga, line 4151: I think the syntax here might be a little wrong because you have a ; at end of second line. I think it is cleaner to declare every variable that is getting a value with a new `let`. I think this syntax right now actually has a bunch of variables technically getting declared as global because of the ; on 4151

PR #286, opened 2020-05-14
enrich data for tracking entries

vsrivastava640:
BTCA-453 - "Enrich Data for Tracking Entries"


lib/datapacksjob.js
```diff
@@ -2961,12 +2963,69 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+
+                    var name=item['Name']
+                    var omniscript_id=item[vloc.namespace+'__OmniScriptId__c'];
+                    var propertyset=JSON.parse(item[vloc.namespace+'__PropertySet__c']);
+                   
+                    records.map(item => {
+                        var element_name= JSON.parse(item[field_name]).ElementName;
+                        var element_omniScript_id= JSON.parse(item[field_name]).OmniScriptId;
+                        var element_type= JSON.parse(item[field_name]).ElementType;
+                        var element_data=JSON.parse(item[field_name])
```
rutlabaga, line 2997: This parse should just be called once and then use the parsed object. Also, please use spaces before and after `=`

lib/datapacksjob.js
```diff
@@ -2961,12 +2963,69 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                var omni_id =records.map(item => {
+                    return JSON.parse(item[field_name]).OmniScriptId;
                 });
 
-                VlocityUtils.log(`${i} of ${totalEntries.totalSize}`);
+                var unique_omni_id=Array.from([...new Set(omni_id)]);
+ 
+                var unique_omni_id_format="";
+                
+                for(var j=0;j<unique_omni_id.length;j++){
```
rutlabaga, line 2975: please use spaces between variables and operators like:
for(var j=0; j < unique_omni_id.length; j++) {

lib/datapacksjob.js
```diff
@@ -2951,6 +2951,8 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
 
         var i = 0;
 
+        var vloc=this.vlocity;
```
rutlabaga, line 2954: Should just set `var self - this;` and use that below.

lib/datapacksjob.js
```diff
@@ -2961,12 +2963,69 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
-                VlocityUtils.log(`${i} of ${totalEntries.totalSize}`);
+                var unique_omni_id=Array.from([...new Set(omni_id)]);
+ 
+                var unique_omni_id_format="";
+                
+                for(var j=0;j<unique_omni_id.length;j++){
+                    if(j == (unique_omni_id.length-1)){
+                        unique_omni_id_format=unique_omni_id_format+`'${unique_omni_id[j]}'`
+                    }else{
+                        unique_omni_id_format=unique_omni_id_format+ `'${unique_omni_id[j]}',`;
```
rutlabaga, line 2979: Will need a max length for this string and to do a loop here or it will fail when over 20000 chars.

lib/datapacksjob.js
vsrivastava640: done, Please find updated code below :- 

                            var element_data = JSON.parse(item[field_name])
                            var element_name = element_data.ElementName;
                            var element_omniScript_id = element_data.OmniScriptId;
                            var element_type = element_data.ElementType;

lib/datapacksjob.js
vsrivastava640: spaces provided between variables and operators

lib/datapacksjob.js
vsrivastava640: done

lib/datapacksjob.js
vsrivastava640: As the limit for characters inside where clause in sosql is 4000, length of "unique_omni_id" is splitted into chunks of 150 to limit length of characters inside sosql where clause.

lib/datapacksjob.js
```diff
@@ -2996,6 +2996,8 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
 
         VlocityUtils.milestone('Downloading Performance Data');
 
+        VlocityUtils.milestone('Downloading Performance Data');
```
rutlabaga, line 2999: This line is duplicated.

lib/datapacksjob.js
```diff
@@ -3016,10 +3020,71 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                            var element_omniScript_id = element_data.OmniScriptId;
+                            var element_type = element_data.ElementType;
+
+
+                            if ((element_name == name) && (element_omniScript_id == omniscript_id)) {
+                                switch (element_type) {
+                                    case "Calculation Procedure Action":
+                                        element_data.configurationName = propertyset.remoteOptions.configurationName;
+                                        break;
+                                    case "DataRaptor":
```
rutlabaga, line 3065: This should be:
case "DataRaptor Extract Action":
case "DataRaptor Post Action":
case "DataRaptor Transform Action":
case "DataRaptor Turbo Action":

lib/datapacksjob.js
```diff
@@ -3016,12 +3018,68 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                    let element_data = JSON.parse(item[field_name])
+                    let element_name = element_data.ElementName;
+                    let element_omniScript_id = element_data.OmniScriptId;
+                    let key = element_name + element_omniScript_id;
+                    let propertyset = JSON.parse(elements[key][self.vlocity.namespace + '__PropertySet__c']);
+                    let element_type = elements[key][self.vlocity.namespace + '__Type__c'];
+
+                    for (let k in propertyset) {
+                        if (keys.includes(k) && (propertyset[k] !== "")) {
+                            element_data[k.substr(k.lastIndexOf('.') + 1)] = propertyset[k]
```
rutlabaga, line 3072: This won't quite work for values with `.` 

If I do:

`var prop = {test: { test2 : "a"}};
console.log(prop['test.test2'])
`

I don't think it will print "a". I think it needs to traverse into the object by the path.

lib/datapacksjob.js
vsrivastava640: removed

lib/datapacksjob.js
```diff
@@ -3016,12 +3018,78 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                    let element_data = JSON.parse(item[field_name])
+                    var element_name = element_data.ElementName;
+                    let element_omniScript_id = element_data.OmniScriptId;
+                    let key = element_name + element_omniScript_id;
+                    let propertyset = JSON.parse(elements[key][self.vlocity.namespace + '__PropertySet__c']);
+                    var element_type = elements[key][self.vlocity.namespace + '__Type__c'];
+
+                    for (let k in propertyset) {
+
+                        if (keys.includes(k) && (propertyset[k] !== "")) {
```
rutlabaga, line 3072: I think this can be changed to `if (k.indexOf('.') != -1)`

rutlabaga: Then you could do k.split('.') then -> propertyset[k][j]. I don't think you need to iterate.

lib/datapacksjob.js
```diff
@@ -3016,12 +3018,78 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                            if (typeof propertyset[k] == 'object') {
+                                for (var j in propertyset[k]) {
+                                    if (keys.includes(`${k}.${j}`) && (propertyset[k][j] !== "")) {
+                                        element_data[j] = propertyset[k][j];
+                                    }
+                                }
+                            } else {
+                                element_data[k] = propertyset[k];
```
rutlabaga, line 3080: Here and above should check that `propertyset[k] != null`.

lib/datapacksjob.js
vsrivastava640: done

lib/datapacksjob.js
vsrivastava640: done

PR #296, opened 2020-07-02
testframework poc

bardmyway:


lib/testframework.js
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    return result;
+};
+
+TestFramework.prototype.executeStart = async function() {
+    var startResult = await this.start(this.jobInfo);
+    let result = await this.resultChunked(startResult, this.jobInfo.resultWithDetails);
+    
+    this.jobInfo.data = {};
```
rutlabaga, line 38: Should remove.

lib/testframework.js
```diff
@@ -0,0 +1,282 @@
...
+    this.jobInfo.data = result;
+    return result;
+}
+
+TestFramework.prototype.executeStartAll = async function() {
+    var startAllResult = await this.startAll(this.jobInfo);
+    let result = await this.resultChunked(startAllResult, this.jobInfo.resultWithDetails);
+    
+    this.jobInfo.data = {};
```
rutlabaga, line 47: Should remove

lib/testframework.js
```diff
@@ -0,0 +1,282 @@
...
+        || !(input.testKeys instanceof Array))
+    {
+        VlocityUtils.error('Start Error', input);
+        return null;
+    }
+
+    let request = {'testKeys' : input.testKeys};
+    VlocityUtils.verbose('Start Request', request);
+
+    let result =  await this.vlocity.jsForceConnection.apex.post('/' + this.vlocity.namespace + '/v1/testframework/start', request, function(err, res) {
```
rutlabaga, line 65: Await will make it return a result, so the callback is not needed.

PR #305, opened 2020-07-27
BTCA-556 Performance data Issue fix

ManasAgarwalVlocity:


lib/datapacksjob.js
```diff
@@ -3086,13 +3086,10 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
 
-                    for (let j = 0; j < chunk.length; j++) {
-                        if (j == (chunk.length - 1)) {
-                            unique_omni_id_format = unique_omni_id_format + `'${chunk[j]}'`
-                        } else {
-                            unique_omni_id_format = unique_omni_id_format + `'${chunk[j]}',`;
-                        }
-                    }
+                    chunk.forEach(omniId => {
+                        unique_omni_id_format += `'${omniId}',`;
```
rutlabaga, line 3090: Probably it would be best to check here that the omniId is not undefined. And also then check that unique_omni_id_format is not "" before running the query

ManasAgarwalVlocity: I thought one replace would be better than 150 if-checks. Will change it.

rutlabaga: My only concern is if they are all undefined then the query will still fail since it will just have `__OmniScriptId__c IN ()` which will throw an exception. It might be good to also add an try catch here to make sure it doesn't fail.

lib/datapacksjob.js
```diff
@@ -3082,24 +3082,28 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
+                        let unique_omni_id_format = "";
 
-                    let chunk = unique_omni_id.splice(0, 150);
-                    let unique_omni_id_format = "";
+                        chunk.forEach(omniId => {
+                            if (omniId) {
+                                unique_omni_id_format += `'${omniId}',`;
+                            }
+                        });
+                        if (unique_omni_id_format) {
```
rutlabaga, line 3094: Wait actually... will this while loop actually finish?

ManasAgarwalVlocity: no... try and catch is inside while loop

PR #311, opened 2020-08-07
Open URL For OS / IP From Process Profiler

vsrivastava640:
BTCA-434

### DIFF
--- a/lib/datapacksjob.js
+++ b/lib/datapacksjob.js
@@ -709,6 +709,7 @@ DataPacksJob.prototype.addRecordToExport = function(jobInfo, queryData, record) 
     record.VlocityDataPackKeyForManifest = queryData.VlocityDataPackType + '/' + (this.vlocity.datapacksexpand.getDataPackFolder(queryData.VlocityDataPackType, record.VlocityRecordSObjectType, record) || record.Id);
     
     if (jobInfo.fullManifest[queryData.VlocityDataPackType][record.Id] || 
+        (jobInfo.specificManifestKeys && queryData.manifestOnly && jobInfo.specificManifestKeys.indexOf(record.VlocityDataPackKeyForManifest) == -1 && jobInfo.specificManifestKeys.indexOf(record.VlocityDataPackKeyForManifest) == -1) ||
         (jobInfo.specificManifestKeys 
         && (!queryData.manifestOnly && jobInfo.specificManifestKeys.indexOf(queryData.VlocityDataPackType) == -1) 
         && jobInfo.specificManifestKeys.indexOf(record.VlocityDataPackKeyForManifest) == -1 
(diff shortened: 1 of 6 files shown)


lib/querydefinition.yaml
```diff
@@ -70,6 +70,10 @@ IntegrationProcedure:
   VlocityDataPackType: IntegrationProcedure
   query: Select Id, %vlocity_namespace%__Type__c, %vlocity_namespace%__SubType__c from %vlocity_namespace%__OmniScript__c WHERE %vlocity_namespace%__IsActive__c
     = true AND %vlocity_namespace%__IsProcedure__c = true
+TestProc:
+  VlocityDataPackType: TestProc
+  query: Select Id, %vlocity_namespace%__Type__c, %vlocity_namespace%__SubType__c from %vlocity_namespace%__OmniScript__c WHERE %vlocity_namespace%__IsActive__c
+    = true AND %vlocity_namespace%__IsProcedure__c = true AND %vlocity_namespace%__IsTest__c = true
 IntegrationProcedureVersion:
   VlocityDataPackType: IntegrationProcedure
   query: Select Id, %vlocity_namespace%__Type__c, %vlocity_namespace%__SubType__c, %vlocity_namespace%__Version__c from %vlocity_namespace%__OmniScript__c WHERE %vlocity_namespace%__IsProcedure__c = true
```
ManasAgarwalVlocity, line 79: `AND %vlocity_namespace%__IsTest__c = false ` might be required to be added for VIP query. Please confirm with @rutlabaga once.

lib/datapacksjob.js
```diff
@@ -3126,12 +3126,12 @@ DataPacksJob.prototype.downloadPerformanceData = async function (jobInfo) {
...
                         element_data.UserProfile = item.CreatedBy.Profile.Name;
-
-                        if (elements[key]) {                        
+                       
+                        if (elements[key]) {                      
                             let propertyset = JSON.parse(elements[key][self.vlocity.namespace + '__PropertySet__c']);
                             var element_type = elements[key][self.vlocity.namespace + '__Type__c'];
 
-                            keys.map(k => {            
+                            keys.map(k => {
```
ManasAgarwalVlocity, line 3134: map should be used when performing action on each element otherwise foreach for normal iteration

vsrivastava640: actually there is no change in this  function(downloadPerformanceData) , its just space diff

lib/querydefinition.yaml
rutlabaga: What is this meant to control? I think better to change the type for TestProc to IntegrationProcedure somewhere in the process.

vsrivastava640: removed query for testproc

lib/datapacksjob.js
```diff
@@ -3291,13 +3291,14 @@ DataPacksJob.prototype.datapackOpen = async function (jobInfo) {
         for (const obj of jobInfo.queries) {
             if (obj.VlocityDataPackType === jobInfo.type) {
                 let query = obj.query.replace(/%vlocity_namespace%__/g, this.vlocity.namespacePrefix);
-                query += query.includes('WHERE') ? ` AND Name = '${jobInfo.name}'` : ` Where Name = '${jobInfo.name}'`;
+                let additionalCondition = `${this.vlocity.namespace}__Type__c = '${jobInfo.omniType}' AND ${this.vlocity.namespace}__SubType__c = '${jobInfo.omniSubType}'`
```
rutlabaga, line 3294: @ManasAgarwalVlocity -- Does htis conflict with your changes? I think the Type / Subtype would only be for OmniScript / IP, so this should be separated by type.

ManasAgarwalVlocity: Yes. If DataPackOpen is still a cli command, not specific to this feature, then it will conflict as it is to open designer by datapack name. Will also require command description update in vlocitycli.js to set input expectations from user.  Also for OmniScript, language may also be required to get unique record.

PR #312, opened 2020-08-07
BTCA-587 Partial Refresh to pull recently created items after last refresh

ManasAgarwalVlocity:


rutlabaga:
This will actually return `No Tracking Entries Found` when there are no new entries. But probably it should only return that if there are also no old entries. Just to make it not seem like an "Error"

PR #321, opened 2020-09-24
PKG-201 Tool for Migrating from Existing DataPacks to new DataPacks for Foundation Package

ManasAgarwalVlocity:


lib/datapacksexpanddefinition.yaml
```diff
@@ -519,6 +556,38 @@ SObjects:
     - DomainObjectAPIName__c
     - DomainObjectFieldAPIName__c
     - InterfaceFieldAPIName__c
+  OmniDataTransformItem__c:
+    DeletedDuringDeploy: true
+    DiffKeys:
+      1: 
+        - MapId__c
```
rutlabaga, line 563: This can be changed to GlocalKey__c

lib/datapacksexpanddefinition.yaml
```diff
@@ -519,6 +556,38 @@ SObjects:
...
+        - InputFieldName__c
+      5: 
+        - FormulaResultPath__c
+      6:
+        - FormulaExpression__c
+        - FormulaSequence__c
+      7:
+        - VlocityRecordSourceKey
+    ListFileName:
+    - _Mappings
```
rutlabaga, line 582: This should be changed to _Items

lib/datapacksjob.js
```diff
@@ -2396,6 +2400,13 @@ DataPacksJob.prototype.deployJob = async function(jobInfo) {
             VlocityUtils.verbose('Getting DataPacks', fullDataPath, tempJobInfo.manifest);
     
             let dataJSON = await this.vlocity.datapacksbuilder.buildImport(fullDataPath, tempJobInfo);
+
```
rutlabaga, line 2403: I think if you want this to fully apply to deploy you will also have to run the updateDataPackWithNewMappings from as part of the deployment itself as well. This would be great to add. buildImport would be a good possible place to put it. But may require other changes (hopefully not though)

lib/datapacksjob.js
```diff
@@ -3354,4 +3358,13 @@ DataPacksJob.prototype.datapackOpen = async function (jobInfo) {
...
         jobInfo.errors.push(err);
     }
+}
+
+DataPacksJob.prototype.convertToCore = async function (jobInfo) {
+    jobInfo.upgradeDataPackFields = true;
+    jobInfo.singleFile = true;
+    let dataJSON = await this.vlocity.datapacksbuilder.buildImport(jobInfo.projectPath, jobInfo);
+
+    let outputPath = jobInfo.outputDir || path.join(jobInfo.tempFolder, 'CoreDataPacks');
```
rutlabaga, line 3368: Should not write to vlocity-temp

lib/datapacksutils.js
```diff
@@ -4917,4 +4917,52 @@ var SALESFORCE_FOLDER_TO_TYPE =  {
...
+                    Object.keys(subObj).forEach(fieldName => {
+                        if (subMappings.fieldNames[fieldName]) {
+                            level2Obj[subMappings.fieldNames[fieldName]] = subObj[fieldName];
+                            if (fieldName !== subMappings.fieldNames[fieldName]) {
+                                delete level2Obj[fieldName];
+                            }
+                        }
+                    });
+                    level2Obj.VlocityRecordSObjectType = childObjDataField;
+                    level2Obj.VlocityRecordSourceKey = level2Obj.VlocityRecordSourceKey.replace(subObj.VlocityRecordSObjectType, childObjDataField);
```
rutlabaga, line 4957: This replace will only replace the frst instance in JS. Needs to use the replace all syntax. `res.replace(new RegExp(subObj.VlocityRecordSObjectType, 'g'), childObjDataField)`

lib/datapacksutils.js
```diff
@@ -4917,4 +4917,52 @@ var SALESFORCE_FOLDER_TO_TYPE =  {
...
+
+    let dataField = this.getDataField(dataPack);
+    let mappings = jobInfo.platformObjectMappings[dataPackType][dataField];
+
+    if (mappings) {
+        let newDataField = mappings.objectName;
+        obj[newDataField] = dataPack.VlocityDataPackData[dataField];
+
+        //Iterate through keys of <dataPack.VlocityDataPackData[dataField]> to update the keys(fieldNames)
+        Object.keys(obj[newDataField][0]).forEach(key => {
```
rutlabaga, line 4934: In this you are missing replacing the VlocityMatchingRecordSourceKey for Element matching keys. It still shows OmniScript__c in the data for the Matching Key.

rutlabaga: There are a few other fields that it needs to be replaced in. If you convert an OmniScript then search through the file for OmniScript__c then it should show all. I just did a find and replace and then it deployed correctly.

PR #322, opened 2020-10-01
Change authToken -> accessToken for oath command in readme

tomcsmith1990:
`authToken` isn't used anywhere else in the code. Based on the Vlocity Build Options, I suspect this should be accessToken. Feel free to correct me if I'm wrong :).


rutlabaga:
Thanks!

PR #328, opened 2020-10-21
BTCA-626 Module Project ItemId__c deployment issue fix

ManasAgarwalVlocity:


lib/datapacksbuilder.js
```diff
@@ -658,6 +658,24 @@ DataPacksBuilder.prototype.getNextImports = function(importPath, jobInfo, curren
 
     var parallelStatus = {};
 
+    var dataPackTypesWithReadyStatus = [];
+    Object.keys(jobInfo.currentStatus).forEach(key => {
+        let type = key.split('/')[0];
+        if ((jobInfo.currentStatus[key] === 'Ready' || jobInfo.currentStatus[key] === 'Added') && !dataPackTypesWithReadyStatus.includes(type)) {
```
rutlabaga, line 664: Add ReadySeparate here too. Is an error state that some can get set to

lib/datapacksbuilder.js
```diff
@@ -658,6 +658,24 @@ DataPacksBuilder.prototype.getNextImports = function(importPath, jobInfo, curren
...
+            dataPackTypesWithReadyStatus.push(type);
+        }
+    });
+
+    var typesToBeDeployedLast = [];
+    var dataPacks = this.vlocity.datapacksutils.dataPacksExpandedDefinition.DataPacks;
```
rutlabaga, line 670: Use the Pattern like:

```
DataPacksUtils.prototype.getPaginationActions = function(dataPackType) {
    return this.getExpandedDefinition(dataPackType, null, "PaginationActions");
}
```

So that it can get the value from the resolved config for each relevant type instead of iterating over the config directly. Probably this can be combined with the above loop to do 1 check for each type and put them in the correct arrays.

PR #331, opened 2020-11-04
OS/VIP/DR migration to STMPA orgs

ManasAgarwalVlocity:


lib/datapacksexpand.js
```diff
@@ -83,7 +83,11 @@ DataPacksExpand.prototype.getNameWithFields = function(nameFields, dataPackData)
         } else if (dataPackData[key] && typeof dataPackData[key] === "string") {
             return dataPackData[key].replace(/[/\\]+/g,' '); // do not allow directory slashes in names
         } else if (dataPackData[key] != null && typeof dataPackData[key] === "object") {
-            return self.getDataPackFolder(null, dataPackData[key].VlocityRecordSObjectType, dataPackData[key]);
+            let sObjDataRecordType = dataPackData[key].VlocityRecordSObjectType;
+            if (jobInfo.upgradeDataPackFields && this.vlocity.datapacksutils.dataPacksExpandedDefinition.StmpaExcludedSObjectsTypes.includes(sObjDataRecordType)) {
```
rutlabaga, line 87: Please use the full same pattern as everything else in dataPacksExpandedDefinition through DataPacksUtils

lib/datapacksexpand.js
```diff
@@ -1077,7 +1094,9 @@ DataPacksExpand.prototype.processDataPackData = async function(dataPackType, par
 
         if (dataPackData.VlocityRecordSObjectType) {
             var sObjectType = dataPackData.VlocityRecordSObjectType;
-
+            if (jobInfo.upgradeDataPackFields) {
```
rutlabaga, line 1097: Shouldn't this only be if it is the right SObject type?

lib/datapacksexpand.js
```diff
@@ -221,6 +225,10 @@ DataPacksExpand.prototype.processList = async function(dataPackType, parentName,
             self.processObjectEntry(dataPackType, dataPack, isPagination);
         });
 
+        if (jobInfo.upgradeDataPackFields) {
```
rutlabaga, line 228: Shouldn't this also be only for certain SObject types?

PR #336, opened 2020-11-17
OS LWC Changes

jfgarcia268:
OS LWC Activation Changes:
- Timeout values
- Change page.waitFor() (Deprecated) to await page.waitForTimeout()
- Log Messages
- Handle particular error when LWC OS Compile failed because of missing custom LWC
- Update jobInfo when LWC Activation fails so the error will show at the end
- Check if Manage Package is Supported

### DIFF
--- a/lib/datapacktypes/omniscript.js
+++ b/lib/datapacktypes/omniscript.js
@@ -159,40 +159,60 @@ OmniScript.prototype.afterActivationSuccess = async function(inputMap) {
             var loginURl = siteUrl + '/secur/frontdoor.jsp?sid=' + sessionToken;
             const browser = await puppeteer.launch(puppeteerOptions);
             const page = await browser.newPage();
+            const loginTimeout = 300000;
 
-            await page.goto(loginURl);
-            await page.waitFor(30000);
+            await Promise.all([
+                page.waitForNavigation({ timeout: loginTimeout, waitUntil: 'load' }),
+                page.waitForNavigation({ timeout: loginTimeout, waitUntil: 'networkidle2'}),
+                page.goto(loginURl, {timeout: loginTimeout})
+            ]);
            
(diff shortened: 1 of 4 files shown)


lib/datapacktypes/omniscript.js
```diff
@@ -2,6 +2,7 @@ var vlocity = require('../vlocity');
 var puppeteer = require('puppeteer-core');
 var fs = require('fs-extra');
 var path = require('path');
+const chromium = require('chromium');
```
rutlabaga, line 5: We won't bundle chromium as it will increase the size of the app too much.

jfgarcia268: Done.. I have removed chromium and Other small changes @rutlabaga

lib/datapacktypes/omniscript.js
```diff
@@ -90,14 +91,28 @@ OmniScript.prototype.afterActivationSuccess = async function(inputMap) {
     var jobInfo = inputMap.jobInfo;
     var dataPack = inputMap.dataPack;
 
+    if(!jobInfo.checkLWCActivation){
```
rutlabaga, line 94: Should only check `if (!jobInfo.ignoreLWCActivation && !jobInfo.checkLWCActivation)`

lib/datapacktypes/omniscript.js
```diff
@@ -90,14 +91,28 @@ OmniScript.prototype.afterActivationSuccess = async function(inputMap) {
...
+        var minVersions = yaml.safeLoad(fs.readFileSync(path.join(__dirname,'..', 'buildToolsMinVersionOS-LWC.yaml'), 'utf8'));
```
rutlabaga, line 95: Should move this to a static variable or to another file so it only loads one.

PR #339, opened 2020-12-15
Deactivate OS when LWC Compile fails - BTCA-673

jfgarcia268:
- Deactivate OS when LWC Compilation Fails (BTCA-673): Line 133 and 273
- Only Check for Supported Manage Package and Retry If OS is LWC

### DIFF
--- a/lib/datapacktypes/omniscript.js
+++ b/lib/datapacktypes/omniscript.js
@@ -96,13 +96,13 @@ OmniScript.prototype.deactivateOmniScript = async function(omniscriptID) {
     try {
         var result = await this.vlocity.jsForceConnection.sobject(osAPIName).update(toUpdate);
         if(!result.success){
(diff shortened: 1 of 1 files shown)


lib/datapacktypes/omniscript.js
```diff
@@ -87,36 +87,57 @@ OmniScript.prototype.createTitleObjects = function(input) {
...
+        }
 
-    if (jobInfo.ignoreLWCActivation) {
-        return;
-    }
+        if (jobInfo.isRetry && jobInfo.omniScriptLwcActivationSkip[dataPack.VlocityDataPackKey]) { 
+            var oldError = jobInfo.omniScriptLwcActivationSkip[dataPack.VlocityDataPackKey];
+            jobInfo.hasError = true;
+            jobInfo.currentStatus[dataPack.VlocityDataPackKey] = 'Error';
+            jobInfo.currentErrors[dataPack.VlocityDataPackKey] = 'LWC Activation Error >> ' + dataPack.VlocityDataPackKey + oldError;
```
rutlabaga, line 131: Probably need a space here in the error message?

jfgarcia268: Agree.. Done

PR #348, opened 2021-01-25
LWC Activation For Classic Cards and Flex Cards

jfgarcia268:


lib/datapacktypes/omniscript.js
```diff
@@ -173,28 +174,33 @@ OmniScript.prototype.afterActivationSuccess = async function(inputMap) {
...
                 await page.waitForTimeout(5000);
 
-                VlocityUtils.log('Starting LWC Compile', omniScriptLogId);
-                
+               
                 let tries = 0;
                 var errorMessage;
-                var maxNumOfTries = jobInfo.defaultMinToWaitForLWC*2;
-                while (tries < maxNumOfTries && !jobInfo.ignoreLWCActivation) {
+                var maxNumOfTries = jobInfo.defaultMinToWaitForLWCOmniScript*2;
```
rutlabaga, line 186: I think we should change to shorter wait between tries and more total tries. 30 seconds is probably too long between checks since the checks are free anyway. My change for this wasn't right to move it to 30 seconds.

jfgarcia268: Agree and Fixed

PR #374, opened 2021-04-29
BTCA-713 Deploy Bug when used along with Expand

ManasAgarwalVlocity:


lib/utilityservice.js
```diff
@@ -386,12 +386,13 @@ UtilityService.prototype.checkLogin = async function() {
...
             } 
 
-            VlocityUtils.verbose('Update Definitions');
-            this.vlocity.datapacksutils.dataPacksExpandedDefinition = this.vlocity.datapacksutils.updateExpandedDefinitionNamespace(this.vlocity.datapacksutils.dataPacksExpandedDefinition);
-
             this.vlocity.PackageVersion = 'No Login';
             this.vlocity.BuildToolSettingVersion = 'latest';
         }
+
+        VlocityUtils.verbose('Update Definitions');
```
rutlabaga, line 393: It is already running this somewhere else. Why is it causing a failure here?

ManasAgarwalVlocity: When using packExpand with packDeploy, files were not getting correctly created.
This functionality was present inside if-block for older namespace fetching code which was getting ignored due to new omniInteractionConfig based code  which returns when namespace is found(for steam orgs)

PR #398, opened 2021-09-21
Adds support for OmniScript LWC compiler

agarcia-sf:


package.json
```diff
@@ -15,6 +15,7 @@
         "js-yaml": "3.13.1",
         "jsforce": "1.9.3",
         "json-stable-stringify": "1.0.1",
+        "runtime-plugin-manager-clone": "^0.1.0",
```
mgoldspink-salesforce, line 18: @agarcia-sf I see this is hosted in your personal github org. Could you please move it to the vlocityinc so that all the dependencies for this that we've forked are in the same org.

mgoldspink-salesforce (changes requested):
@agarcia-sf Can you please move the

package.json
agarcia-sf: Yeah!, will do... this is temporary while I work on the PR to the original repo.

lib/buildToolsMinVersionOS-LWC.yaml
```diff
@@ -1,2 +1,3 @@
 vlocity_cmt: '900.374'
-vlocity_ins: '890.220'
\ No newline at end of file
+vlocity_ins: '890.220'
+omnistudio: '236.0.0'
```
rutlabaga, line 3: I don't think this is right since this also affects the other compile.

lib/datapacktypes/omniscript.js
```diff
@@ -154,11 +154,11 @@ OmniScript.prototype.afterActivationSuccess = async function (inputMap) {
             return;
         }
 
-        await this.compileOSLWC(jobInfo, omniScriptId,omniScriptKey);
-        
+        await this.compileOmniScriptLwc(jobInfo, omniScriptId, omniScriptKey);
```
rutlabaga, line 157: It can't replace the old way. It should still use old way if not at the right version for this.

lib/datapacksjob.js
```diff
@@ -2647,6 +2647,28 @@ DataPacksJob.prototype.deployJob = async function(jobInfo) {
...
+                sourcepath: tempSfdxFolder,
+                targetusername: this.vlocity.sfdxUsername
+            };
+
+            // Ensure there is a sfdx-project.json file
+            this.vlocity.datapacksutils.getSFDXProject({...jobInfo, sfdxForcePath: tempSfdxFolder});
+
+            // Deploy using sfdx
+            const deployGeneratedLwcFolder = await this.vlocity.utilityservice.sfdx('source:deploy', tempDeployOptions);
+            jobInfo.currentStatus = Object.assign(jobInfo.currentStatus, this.vlocity.datapacksutils.formatSalesforceDeployResponse(deployGeneratedLwcFolder));
```
rutlabaga, line 2666: Would be good to add a test for pass and fail. Can discuss.

package.json
rutlabaga, line 18: Can you remove the ^

agarcia-sf: done!

Issue #400, opened 2021-10-07
Puppeteer over a proxy


MaciekLeks:
Does the tool support proxy for Puppeteer? If yes, what to do to make it works?

MaciekLeks:
The simplest answer after digging in the code is NOT. Working behind the proxy and having puppeteer on the board makes the tool to crazy. 

Example log:
```
Sample error log part for Chromium without --proxy-server behind a proxy:

 Migrated 2333 items >> Deploying  
  Starting LWC Activation for Classic Card >> VlocityUILayout/b2bSampleAppCard  
  Error: net::ERR_TUNNEL_CONNECTION_FAILED at https://productivity-leo-1124.cs101.my.salesforce.com/apex/vlocity_cmt__carddesignernew?id=a5t1X000000lgZNQAY&compile=true
    at navigate (/usr/local/lib/node_modules/vlocity/node_modules/puppeteer-core/lib/cjs/puppeteer/common/FrameManager.js:115:23)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async FrameManager.navigateFrame (/usr/local/lib/node_modules/vlocity/node_modules/puppeteer-core/lib/cjs/puppeteer/common/FrameManager.js:90:21)
    at async Frame.goto (/usr/local/lib/node_modules/vlocity/node_modules/puppeteer-core/lib/cjs/puppeteer/common/FrameManager.js:417:16)
    at async Page.goto (/usr/local/lib/node_modules/vlocity/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Page.js:784:16)
    at async module.exports.VlocityUILayout.onDeployFinish (/usr/local/lib/node_modules/vlocity/lib/datapacktypes/vlocityuilayout.js:75:21)
    at async module.exports.VlocityCard.onDeployFinish (/usr/local/lib/node_modules/vlocity/lib/datapacktypes/vlocitycard.js:167:9)
    at async module.exports.DataPacksUtils.handleDataPackEvent (/usr/local/lib/node_modules/vlocity/lib/datapacksutils.js:2816:10)
    at async module.exports.DataPacksJob.deployJob (/usr/local/lib/node_modules/vlocity/lib/datapacksjob.js:2630:13)
    at async module.exports.DataPacksJob.doRunJob (/usr/local/lib/node_modules/vlocity/lib/datapacksjob.js:337:9)  
[...]

PR #401, opened 2021-10-09
Chromium proxy server set with sf.httpProxy

MaciekLeks:
Applying --proxy-server into Puppeteer to make the tool work properly behind a proxy server. Without it sfdx part would work properly with a proxy and only a part of a tool will not work and forced the tool to raise net::ERR_TUNNEL_CONNECTION_FAILED error.

Related issue: https://github.com/vlocityinc/vlocity_build/issues/400


BobSeu:
Out of curiosity, how long does it usually take to go from alpha to the main branch?

Issue #416, opened 2021-11-25
1.14.14 proxy fix broke omniscript activation on gitlab pipeline


KevinGossentCap:
the way it's done in fix #400 seem to break omniscript activation from our on-premise gitlabci pipeline (it was previously working)

KevinGossentCap:
Just tested, it's solved with #417 
:)
And proxy still working (proxy of #414  solved my local deployment issue and still working)

PR #417, opened 2021-11-25
other puppeteer proxy handling

KevinGossentCap:
try making sure the puppeteer proxy-server is set only if jobInfo.httpProxy exist
might solve #416


oxycash:
@KevinGossentCap does this mean the job yaml need to have httpProxy: proxy:port info to have puppeteer run over a proxy?

PR #419, opened 2021-12-01
Change Job Option from httpProxy to puppeteerHttpProxy

jfgarcia268:
Change Job Option from httpProxy to puppeteerHttpProxy to avoid confusion with sf.httpProxy and also include it in VLOCITY_COMMANDLINE_OPTIONS


oxycash:
@jfgarcia268 what's the syntax to specify httpproxy?

PR #433, opened 2022-01-24
gitCheck Support  for OmniStudio with job setting "generalSettingsCus…

jfgarcia268:
gitCheck Support  for OmniStudio with job setting "generalSettingsCustomSetting"

### DIFF
--- a/lib/utilityservice.js
+++ b/lib/utilityservice.js
@@ -362,14 +362,14 @@ UtilityService.prototype.sfdxLogin = async function() {
     }
 }
 
-UtilityService.prototype.checkLogin = async function() {
-    
+UtilityService.prototype.checkLogin = async function(options) {
+
     VlocityUtils.verbose('Check Login');
     try {
         if (this.vlocity.sessionId || this.vlocity.accessToken) {
             await this.getNamespace();
             await this.checkOmniStudioPackage();
-            await this.checkRequiredVersion();
+            await this.checkRequiredVersion(options);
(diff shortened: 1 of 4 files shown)


lib/utilityservice.js
```diff
@@ -764,22 +764,28 @@ UtilityService.prototype.getDataPackKeyFromFilename = function(filename, jobInfo
     return null;
 }
 
-UtilityService.prototype.setVlocitySetting = async function(settingName, value) {
+UtilityService.prototype.setVlocitySetting = async function(settingName, value, jobInfo) {
+    var customSettingAPIName = `${this.vlocity.namespacePrefix}GeneralSettings__c`;
+    var valueColumn = `${this.vlocity.namespacePrefix}Value__c`;
+    if (jobInfo.generalSettingsCustomSetting) {
+        customSettingAPIName = jobInfo.generalSettingsCustomSetting
+        valueColumn = 'Value__c';
```
rutlabaga, line 772: What is this value column? Is this meant to be for OmniStudio OmniInteractionConfig? It should be Value if so. And I am not sure Set will work for this without using the tooling API since it moved to Metadata.

jfgarcia268: valueColumn is the API name of the Customsetting Colum that has the Actual value of the setting. The idea is that The customer creates a new CustomSetting with a column named 'Value__c'. The new Custom Setting API Name will be added by the customer tp the jobfile using "generalSettingsCustomSetting:<xxx>"

PR #482, opened 2022-06-21
Added the functionality to let user override the org commit for diff via job file

92SGupta:
There are instances where we need to quickly redeploy some commit from local, to do so we need to login to the org and modify the VBTDeploykey. So instead of doing that user can add the "**overrideOrgCommit: local_commit_ID**" to let the VBT perform the git diff on the same and then deploy the changes. Once deployed, the HEAD commit of the branch would be updated in the target org as usual.

### DIFF
--- a/lib/vlocitycli.js
+++ b/lib/vlocitycli.js
@@ -106,7 +106,8 @@ var VLOCITY_COMMANDLINE_OPTIONS = {
     "puppeteerHeadless": Boolean,
     "ignoreLWCActivationOS": Boolean,
     "ignoreLWCActivationCards": Boolean,
-    "puppeteerHttpProxy": String
+    "puppeteerHttpProxy": String,
+    "overrideOrgCommit": String
 };
 
 var VLOCITY_COMMANDLINE_OPTIONS_SHORTHAND = {


lib/utilityservice.js
```diff
@@ -652,12 +652,21 @@ UtilityService.prototype.getGitChangesFromCommand = async function(gitChanges, a
 UtilityService.prototype.getGitDiffsFromOrgToLocal = async function(jobInfo) {
     try {
         var vbtDeployKey = `VBTDeployKey${jobInfo.gitCheckKey ? jobInfo.gitCheckKey : ''}`;
-        var hashKey = await this.getVlocitySetting(vbtDeployKey);
```
rutlabaga, line 655: can you change this to let instead of two var declarations.

rutlabaga:
Please also add details to readme. This is a great idea.

lib/utilityservice.js
92SGupta: @rutlabaga: Can you check it now, if it's fine.

92SGupta:
> Please also add details to the readme. This is a great idea.

Done

92SGupta:
> overrideOrgCommit

I have added the option now @jfgarcia268 @rutlabaga

PR #484, opened 2022-06-30
Add CreatedBy. CreatedDate, LastModified and LastModifiedDate to packGetAllAvailableExports

jfgarcia268:
Add the following to packGetAllAvailableExports:
1. CreatedByName 
2. CreatedDate 
3. LastModifiedName
4. LastModifiedDate

### DIFF
--- a/lib/datapacksjob.js
+++ b/lib/datapacksjob.js
@@ -804,6 +804,20 @@ DataPacksJob.prototype.addRecordToExport = function(jobInfo, queryData, record) 
     record.VlocityDataPackType = queryData.VlocityDataPackType;
     record.VlocityRecordSObjectType = record.attributes.type;
 
+    if(record.CreatedBy){
+        if(record.CreatedBy.Name){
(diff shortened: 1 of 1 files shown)


lib/datapacksjob.js
```diff
@@ -851,6 +865,9 @@ DataPacksJob.prototype.runQueryForManifest = async function(queryInput) {
         }
 
         var query = this.vlocity.omnistudio.updateQuery(queryData.query).replace(/%vlocity_namespace%__/g, this.vlocity.namespacePrefix);
+        query = query.replace("from", "FROM");
+        const subQuery = query.split('FROM');
```
rutlabaga, line 869: This will need just a little extra logic. It should be `replace(" from ", " FROM ")` and split on " FROM " only if the query contains " FROM " since right now it will just break queries which contain "FRom" or "fromDate".

jfgarcia268: @rutlabaga Done. Thanks

PR #487, opened 2022-07-18
hotfix deploy generated lwc to prod

AllanOricil:
This fix enables vlocity cli to deploy synthetized LWCs to Prod without specifying any tests. 

This issue is mentioned here:
https://salesforce.stackexchange.com/questions/271433/lwc-without-apex-sfdx-forcesourcedeploy-into-production

It fixes this issue using `sfdx force:mdapi:deploy` instead of `sfdx force:source:deploy`


lib/datapacksjob.js
```diff
@@ -2727,28 +2730,61 @@ DataPacksJob.prototype.deployJob = async function(jobInfo) {
...
-            let tempDeployOptions = {
-                sourcepath: tempSfdxLwcFolder,
-                targetusername: this.vlocity.sfdxUsername
-            };
-
-            // Ensure there is a sfdx-project.json file
-            this.vlocity.datapacksutils.getSFDXProject({...jobInfo, sfdxForcePath: tempSfdxLwcFolder});
-            try {
-                // Deploy using sfdx
-                const deployGeneratedLwcFolder = await this.vlocity.utilityservice.sfdx('source:deploy', tempDeployOptions);
```
rutlabaga, line 2742: Thanks for this! These true production environment use cases are harder for us to test since we primarily use Developer Environments that don't have full deployment protection, which is a bit of a gap...

But, instead of fully removing the flow here, can you let it run instead if  `if(!commandExistsSync('sfdx')) {`

Right now this uses the `salesforce-alm` module directly through this utility service - `UtilityService.prototype.sfdx`. This does have some small benefits that the Desktop App user doesn't have to install SFDX CLI if they are unfamiliar with it. Supporting using sfdx cli directly is a great idea though.

AllanOricil: @rutlabaga my first implementation used the`salesforce-alm` package, but I was having a null pointer exception in this line:

https://github.com/forcedotcom/salesforce-alm/blob/a83cd97149dc2b479813619410730a190de4c962/src/ToolbeltCommand.ts#L442

the problem is that `this.ux` is `null`. I couldn't understand the reason for that. Would you know?

rutlabaga: You just have to initialize it like:  `sfdxCommand.ux = await require("@salesforce/command").UX.create();` 

I like your approach actually and would want to switch the existing `UtilityService.prototype.sfdx` to use either/or depending on job settings since it would be more flexible. But then all the commands need to be implemented to use both which could take some effort.

AllanOricil: I can create a module that can receive a job prop to either use sfdx or salesforce-alm. When given a job prop like `useClientSfdx` || `useHostSfdx`, this module runs sfdx commands using the sfdx found in the host machine, but only if it exists. When sfdx does not exist or `useClientSfdx` is false, this module defaults to salesforce-alm. Then we can use this module as a service everywhere it is needed. What do you think? The api would be sfdx.<Command in camel case>, like sfdx.mdapiDeploy(flags).

AllanOricil: But I would also take the opportunity to ask what do u think about rewriting this cli as a sfdx plugin?  The idea is that this repo could be a module under @salesforce or @vlocity namespace, that provides services to process vlocity datapacks. Then we could create a sfdx plugin to use services provided by this module. We could even make other tools use this module as a service, such as the vlocity vscode extension or vlocity dx ui.
I have some experience with sfdx/ocliff plugins framework and I could help with that. I can't really estimate how long it would take to complete this rework, because I didn't really use this cli, or vlocity, that much, but I believe we could release this plugin in pieces, where each vlocity command would be a piece.
Would this be something salesforce/vlocity would like to invest time and money?

AllanOricil:
@rutlabaga  

I added `useHostSfdx` to enable developers to use their own instance of sfdx. 

````yaml
...
useHostSfdx: true
...
```

I have also added a method to select between the sfdx in the host machine and salesforce-alm commands. 

````js
const sfdxCommandRunner = this.vlocity.utilityservice.getSfdxCommandRunner(jobInfo);

await sfdxCommandRunner(command, options);
````

**Comment by AllanOricil** [comment]
@rutlabaga should I rebase on top of `vlocityinc:beta` or is it fine to be use `vlocityinc:master`?

File: lib/utilityservice.js [hunk 2]
```diff
@@ -874,6 +886,18 @@ UtilityService.prototype.sfdx = async function(command, options) {
         } else if (command == 'mdapi:listmetadata') {
             var MdapiListmetadataCommand = require("salesforce-alm/dist/commands/force/mdapi/listmetadata");
             sfdxCommand = new MdapiListmetadataCommand.MdapiListmetadataCommand();
+        } else if (command == 'source:convert') {
+            var SourceConvertCommand = require("salesforce-alm/dist/commands/force/source/convert");
+            flags.quiet = true;
```
rutlabaga, line 891: For these `flags.` you can remove since they are set above. The one for 'org:display' is always set to quiet because it can display a token that shouldn't be there in the logs.

lib/utilityservice.js
```diff
@@ -834,6 +836,16 @@ UtilityService.prototype.getVlocitySetting = async function(settingName) {
     return null;
 }
 
+UtilityService.prototype.getSfdxCommandRunner = function(jobInfo) {
+    if(jobInfo.useHostSfdx === true && commandExistsSync('sfdx')) {
```
rutlabaga, line 840: Please change to `useSfdxCli`.

AllanOricil: done

lib/utilityservice.js
AllanOricil: both `flags.quiet` and `flags.json` were removed

AllanOricil:
@rutlabaga I did the changes you asked. Is there anything else you would like me to change?

PR #534, opened 2022-12-01
Fixes an issue with exporting Products, etc with hyphens.

lmckusick:
This fixes an issue with exporting a String Translation (and possibly other items), where the Name is a single hyphen. The data sanitization removes the hyphen, leaving the resulting folder name as an empty string, and breaking the export. This ensures that if the sanitization removes the full string, that there is a replacement.

### DIFF
--- a/lib/datapacksexpand.js
+++ b/lib/datapacksexpand.js
@@ -31,6 +31,10 @@ DataPacksExpand.generateFolderOrFilename = function(filename, extension) {
         .replace(/^[-_\\/]+/, "")
         .replace(/[-_\\/]+$/, "");
 
+    if (sanitizedFilename == '') {
+        sanitizedFilename = 'MissingName';
+    }
+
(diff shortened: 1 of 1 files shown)


twu-sf (approved):

agarcia-sf (approved):

mgoldspink-salesforce (approved):

lib/datapacksexpand.js
```diff
@@ -31,6 +31,10 @@ DataPacksExpand.generateFolderOrFilename = function(filename, extension) {
         .replace(/^[-_\\/]+/, "")
         .replace(/[-_\\/]+$/, "");
 
+    if (sanitizedFilename == '') {
+        sanitizedFilename = 'MissingName';
```
manas-sf, line 35: We can return something similar like line#22 -- return 'MissingName-' + datapackutils.guid(); And also add a log that why it's missing  the name.

PR #562, opened 2023-02-21
W-12484068: Convert Batch Deployment to Use CSV instead of raw JSON

lmckusick:
This change converts the data sent during a batch deployment to CSV format instead of passing raw JSON.

This change originated due to an issue with deploying huge Calculation Matrices, where the row data file hit the 10m character limit of the Bulk API.

This also adjusts permissions on the TESTVDXProfile Profile used in runTestJob, adding some missing permissions and resolving some issues found while running the test job.


mgoldspink-salesforce (changes requested):
@lmckusick I think we need one more change here (which is totally unrelated to what you've done). 

I can see CodeShip builds failing and it's because it's using a super old version of node (v10 - https://github.com/vlocityinc/vlocity_build/blob/master/Dockerfile).

Could you please change that to 16 and let's see if that helps the build pass again.

mgoldspink-salesforce (approved):

package.json
```diff
@@ -16,12 +16,13 @@
         "js-yaml": "3.13.1",
         "jsforce": "1.9.3",
         "json-stable-stringify": "1.0.1",
-        "runtime-plugin-manager-clone": "0.1.0",
+        "json2csv": "^6.0.0-alpha.2",
```
manas-sf, line 19: Should we remove ^ ?

manas-sf (approved):

package.json
mgoldspink-salesforce: Agreed - let's lock it to the version we test with

lmckusick: Fixed.

agarcia-sf (approved):

PR #571, opened 2023-03-03
W-12484068 Split bulkRecordsData into small batches to prevent exceed…

manas-sf:
…ing character limit

### DIFF
--- a/lib/datapacksjob.js
+++ b/lib/datapacksjob.js
@@ -2301,12 +2301,38 @@ DataPacksJob.prototype.deployBulkRecords = async function (dataPack, BulkRecords
         BulkRecords = JSON.parse(JSON.stringify(BulkRecords).replace(/%vlocity_namespace%__/g, ''));
     }
 
-    var result = await self.vlocity.datapacksutils.createBulkJob(objectName, 'insert', BulkRecords, randomNum);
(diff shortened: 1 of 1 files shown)


lib/datapacksjob.js
```diff
@@ -2301,12 +2301,38 @@ DataPacksJob.prototype.deployBulkRecords = async function (dataPack, BulkRecords
         BulkRecords = JSON.parse(JSON.stringify(BulkRecords).replace(/%vlocity_namespace%__/g, ''));
     }
 
-    var result = await self.vlocity.datapacksutils.createBulkJob(objectName, 'insert', BulkRecords, randomNum);
+    const BATCH_SIZE_FOR_BULK_JOB = 1000;
```
mgoldspink-salesforce, line 2304: Is there a reason to choose 1000? Could we be smarter about how we pick this number and avoid batching if we don't need to?

manas-sf: @mattgoldspink Yes. I was thinking of checking if it's more than 5~6k, then only we can split.
Also we can increase the count to 4k or above.

Also we can have input from user itself, that they can provide to paginate.

This is just POC and open to improvements. Thanks

manas-sf: @lmckusick FYI

Issue #475, opened 2022-06-13
FlexCard Compilation Error


lotusboy:
Hi,

As part of our vlocity deployment, we have 3 FlexCards which are failing to compile and so cannot subsequently be deployed. Omniscripts are compiling successfully.

Interestingly, if I set lwcCompilerVersion: "890.331", meaning the legacy method is used, the FlexCards get compiled and deployed successfully.

Vlocity Build v1.15.3
Vlocity version 890.331

Log showing the first FlexCard error:
...
  Loading @vlocity-ins/flexcard-compiler@890.331  
  Compiling Flexcard with  FlexCard ID:  >> a6J0E000002GcOkUAK  
  /usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginVm.js:74
            throw new Error("Invalid javascript file " + filePath);
            ^

Error: Invalid javascript file /Users/sloftus/Documents/git/smartchassis/vlocity-temp/plugins/copy-anything/dist/index.cjs
    at PluginVm.load (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginVm.js:74:19)
    at PluginManager.load (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginManager.js:553:24)
    at PluginManager.require (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginManager.js:191:21)
    at PluginVm.sandboxRequire (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginVm.js:252:33)
    at Object.assign.resolve (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginVm.js:178:25)
    at /Users/sloftus/Documents/git/smartchassis/vlocity-temp/plugins/less/lib/less/utils.js:8:23
    at /Users/sloftus/Documents/git/smartchassis/vlocity-temp/plugins/less/lib/less/utils.js:127:2
    at Script.runInContext (node:vm:139:12)
    at PluginVm.vmRunScriptInSandbox (/usr/local/lib/node_modules/vlocity/node_modules/runtime-plugin-manager-clone/dist/src/PluginVm.js:136:16)
[...]

PR #577, opened 2023-03-07
Move env var, so it's compatible with Windows

thvd:
Should fix: https://github.com/vlocityinc/vlocity_build/issues/475#issuecomment-1458586636

### DIFF
--- a/lib/sfdx.js
+++ b/lib/sfdx.js
@@ -1,8 +1,8 @@
 const shell = require('shelljs');
 
 const _run = (command) => {
-	command = 'FORCE_COLOR=0 ' + command + ' --json';
-	const exec = shell.exec(command, { silent: true });
+	command += ' --json';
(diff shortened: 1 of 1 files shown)


KevinGossentCap (commented):
be carefull, you have to add that to the env option array: ...process.env,
else you remove all use of system env var

KevinGossentCap:
the line should be like 
```
const exec = shell.exec(command, { env: { ...process.env, FORCE_COLOR:0, }, silent: true });
```
I've tested it locally and it works :)

thvd:
Should be good to go now. @KevinGossentCap thanks for testing

KevinGossentCap (approved):

thvd:
@manas-sf  do you have an ETA when this patch will be release to stable version? Because now the latest version 1.17.1 isn't working for Windows users of my team

KevinGossentCap:
@manas-sf As I've commented on #581 , you produced 1.71.2 tag but the version have not been released at all...
See recent issue #584

PR #604, opened 2023-08-18
Add %vlocity_namespace%__CatalogCode__c to Catalog Query

jfgarcia268:
@manas-sf:
Add %vlocity_namespace%__CatalogCode__c to Catalog Query so it can be used as VlocityDataPackKey


manas-sf:
Will be delivered through https://github.com/vlocityinc/vlocity_build/pull/607

PR #636, opened 2024-01-23
 updated version

ABHINAVKR:


package.json
```diff
@@ -2,7 +2,7 @@
     "author": "arutland@vlocity.com",
     "bin": "lib/vlocitybuild.js",
     "dependencies": {
-        "async": "3.2.0",
+        "async": "^3.2.5",
```
manas-sf, line 5: Please remove ^ as we don't want any unknown update.

package.json
```diff
@@ -23,13 +23,13 @@
         "properties": "1.2.1",
         "puppeteer-core": "^5.3.1",
         "runtime-plugin-manager-clone": "0.1.0",
-        "salesforce-alm": "49.5.0",
+        "salesforce-alm": "54.8.5",
```
manas-sf, line 26: Did we try running npm install and doing packExport or some tests?

ABHINAVKR: packExport worked, will do other tests

PR #716, opened 2025-03-17
Vulnarebility and sf changes

mdmehran-qureshi:


codeship-steps.yml
```diff
@@ -3,5 +3,5 @@
   command: ./codeship/test.sh
 - name: publish
   service: vlocitybuild
-  tag: ^(master|alpha|beta)
+  tag: ^(master|alpha|beta|alpha-12)
```
sf-pranav-varshney, line 6: This is not required.

sf-pranav-varshney (approved):

codeship-steps.yml
mdmehran-qureshi: Updated

PR #795, opened 2026-05-29
Upgrade puppeteer-core from v6 to v24

gabrielcowley:
Bumps `puppeteer-core` from `^6.0.0` (Sep 2020) to `^24.43.0`, an 18-major-version jump, and replaces the 9 call sites that use `page.waitForTimeout()` (removed in Puppeteer v22.0.0).

### Why v24 specifically, not the latest v25

`puppeteer-core@25.x` requires Node `>=22.12.0` (v25 dropped Node 18 - see [puppeteer/puppeteer#14342](https://github.com/puppeteer/puppeteer/issues/14342)). This project's `package.json` declares `engines.node: ">=18.8.2"`, so jumping to v25 would silently raise that floor and drop Node 18/20 users of the published `vlocity` CLI. `puppeteer-core@24.43.x` is the newest release line that still declares `engines.node: ">=18"`, keeping the existing Node support contract intact.

### Required code changes

`page.waitForTimeout()` was removed in Puppeteer v22.0.0. The 9 call sites across the four DataPack-type modules are rewritten to the same `setTimeout`-based idiom that's already used elsewhere in these files for `msToWaitForLWCDeploy`.

No other Puppeteer API changes were required - `puppeteer.launch`, `browser.newPage`, `page.goto`, `page.waitForNavigation`, `page.waitForSelector`, `page.evaluate`, `browser.close`, and the `headless` / `executablePath` / `args` launch options used here are all still supported in v24.

### Behavioural changes from the upgrade

Two pre-existing things worth being aware of when reviewing - neither is changed by this PR, but both are touched by the upgrade in practice:

1. `headless: true` semantics changed in Puppeteer v22. It now launches Chrome's new headless mode (real Chrome) rather than the old shell. With the system-Chrome executablePath selected by `UtilityService.getPuppeteerOptions`, this should work the same, but it is a behavioural change in the underlying browser.
2. The `.local-chromium` fallback in `lib/utilityservice.js` is effectively dead code with modern Puppeteer. It scans `node_modules/puppeteer/.local-chromium` with `chrome-mac` / `chrome-linux` / `chrome-win` folder names. Since Puppeteer v15 the bundled Chromium has been managed by `@puppeteer/browsers`, which uses a different cache layout (chrome-mac-x64 / chrome-linux64 / chrome-win64) in a different directory. This path matters only for users who relied on `npm install puppeteer -g` to provide the browser - anyone using system Chrome or `puppeteerExecutablePath` is unaffected.

### DIFF
--- a/lib/datapacktypes/flexcard.js
+++ b/lib/datapacktypes/flexcard.js
@@ -116,7 +116,7 @@ FlexCard.prototype.monitorFlexCardCompilation = async function(page, jobInfo, co
             errorMessage = ' Error: ' + e;
         }
         tries++;
-        await page.waitForTimeout(jobInfo.defaultLWCPullTimeInSeconds * 1000);
+        await new Promise(resolve => setTimeout(resolve, jobInfo.defaultLWCPullTimeInSeconds * 1000));
     }
 
     if (tries == maxNumOfTries) {
@@ -182,7 +182,7 @@ FlexCard.prototype.compileBulkFlexCards = async function(jobInfo, idsArray, page
     VlocityUtils.verbose('LWC FlexCards Activation URL', flexCardCompilePage);
 
     await page.goto(flexCardCompilePage, {timeout: loginTimeout});
-    await page.waitForTimeout(5000);
+    await new Promise(resolve => setTimeout(resolve, 5000));
 
     // Use common monitoring method
     const context = {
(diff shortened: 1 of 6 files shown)


gabrielcowley:
Unfortunately, I can't see the build failure details and I'm not sure how to get `npm run test` working locally. All 46 unit tests pass when I run `npm run unitTest`.

PR #798, opened 2026-06-11
@W-22360038 Fixing npm audit vulnerabilties

sf-pallavi-das:
## Summary
- Adds `tar` override (`^7.5.16`) under `runtime-plugin-manager-clone` to patch 6 high-severity advisories with no available v6 fix
- `npm audit` now reports **0 vulnerabilities** (was 2 high)

## Why this is needed
The audit findings are 6 advisories against `tar@6.x`, all coming through `runtime-plugin-manager-clone`. `tar` v6 has no patched release; v7 is the only available fix. Scoping the override under `runtime-plugin-manager-clone` keeps the change minimal.

## Compatibility
`runtime-plugin-manager-clone` calls `tar.extract({file, cwd, strip})` — that signature is unchanged in v7. v7's CommonJS export still resolves via `require('tar')`. Engines field already requires Node `>=18.8.2`, satisfying tar v7.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,33 +49,30 @@
             }
         },
         "node_modules/@babel/runtime": {
-            "version": "7.28.4",
-            "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.28.4.tgz",
-            "integrity": "sha512-Q/N6JNWvIvPnLDvjlE1OUBLPQHH6l3CltCEsHIujp45zQUSSh8K+gHnaEX45yAT1nyngnINhvWtzN+Nb9D8RAQ==",
-            "license": "MIT",
+            "version": "7.29.7",
+            "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.29.7.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -583,14 +583,12 @@
         "node_modules/@types/semver": {
             "version": "7.7.1",
             "resolved": "https://registry.npmjs.org/@types/semver/-/semver-7.7.1.tgz",
-            "integrity": "sha512-FmgJfu+MOcQ370SD0ev7EI8TlCAfKYU+B4m5T3yXc1CiRN94g/SZPtsCkk506aUDtlMnFZvasDwHHUcZUEaYuA==",
-            "license": "MIT"
```
mdmehran-qureshi, line 587: Why did it remove license from all dependencies?

sf-pallavi-das: Updated

sf-pranav-varshney (approved):

mdmehran-qureshi (approved):
lgtm

PR #802, opened 2026-09-10
@W-24123570: stop logging live session token via frontdoor URL

snehaljha-sf:
VlocityUtils.verbose('LWC Activation Login URL', loginURl) logged the full /secur/frontdoor.jsp?sid=<accessToken> URL, including the live jsForce session token, whenever --verbose is enabled. That message ends up in VlocityUtils.fullLog, which is written to VlocityBuildLog.yaml and a copy under <tempFolder>/logs/ - a persisted plaintext file, not just console output. Anyone able to read that file/CI artifact could reuse the token via frontdoor.jsp to hijack the runner's authenticated session (CWE-532/598).

loginURl is only needed for page.goto() right after, matching the sibling LWC-activation call sites in flexcard.js/vlocitycard.js/ vlocityuilayout.js, which build the same frontdoor URL but never log it.

### DIFF
--- a/lib/datapacktypes/omniscript.js
+++ b/lib/datapacktypes/omniscript.js
@@ -315,7 +315,6 @@ OmniScript.prototype.compileOSLWC = async function (jobInfo, omniScriptId, omniS
                     var siteUrl = this.vlocity.jsForceConnection.instanceUrl;
                     var sessionToken = this.vlocity.jsForceConnection.accessToken;
                     var loginURl = siteUrl + '/secur/frontdoor.jsp?sid=' + sessionToken;
-                    VlocityUtils.verbose('LWC Activation Login URL', loginURl);
                     var browser;
     
                     try {


lib/datapacktypes/omniscript.js
```diff
@@ -315,7 +315,6 @@ OmniScript.prototype.compileOSLWC = async function (jobInfo, omniScriptId, omniS
                     var siteUrl = this.vlocity.jsForceConnection.instanceUrl;
                     var sessionToken = this.vlocity.jsForceConnection.accessToken;
                     var loginURl = siteUrl + '/secur/frontdoor.jsp?sid=' + sessionToken;
-                    VlocityUtils.verbose('LWC Activation Login URL', loginURl);
```
sf-pallavi-das, line 318: @snehaljha-sf , 
Customers used this url to manually activate the components for cases where puppeteer/local compiler fails to do so. It makes sense from security perspective but do you think we can provide any alternative?

snehaljha-sf: @sf-pallavi-das , As discussed, since it's not a documented way and is a security vulnerability, Can you please approve

sf-pallavi-das (approved):

PR #807, opened 2026-09-15
@W-24098247@ Escape QueryService SOQL WHERE-clause values to prevent SOQL injection (CWE-89)

amaindoliya-sf:
### Problem
`QueryService.buildWhereClauseValueByType()`'s default (text) branch built
`fieldName + " = '" + value + "'"` with no escaping. A value containing a single
quote (e.g. a crafted DataPack field like `' OR 1=1 --`) breaks out of the SOQL
string literal, allowing injection under the authenticated connection's permissions.

The finding also named `lib/datapacktypes/omniscript.js`. Those `searchKey`/Type/
SubType/Language sites were fixed on `alpha` by #805 (`@W-24123452@`), but with a
**module-local** `escapeSOQLValue` helper duplicated inside that file.

### Fix
- `lib/utilityservice.js`: new shared `escapeSOQLString(value)` — escapes `\` first,
  then `'` — so any query builder can reuse it (the finding recommended a centralized
  escaping utility).
- `lib/queryservice.js`: the text branch now escapes `value` via `escapeSOQLString`.
- `lib/datapacktypes/omniscript.js`: **consolidation** — removed the duplicate
  module-local `escapeSOQLValue` and pointed its call sites at the shared
  `this.vlocity.utilityservice.escapeSOQLString`. One escaper, one place to reason
  about. Behavior is preserved: those values are interpolated inside template
  literals, so `null`/`undefined` still coerce to `'null'`/`'undefined'` exactly as
  before.

### Tests
* `test/utilityservice.spec.js`: unit tests for `escapeSOQLString` (single quote,
backslash, backslash-before-quote, quote-breakout payload, ordinary value,
null/undefined) plus a `QueryService.buildWhereClauseValueByType` test asserting a
breakout payload is neutralized. The old `test/omniscript.spec.js` only tested the
now-removed local helper; its distinctive cases were folded into the shared spec and
the file removed. `npm run unitTest` → all green (54 passing).
* Cloudbee test success link: https://cloudbees.io/vlocityinc/components/670f856a-778d-4d66-9417-519d4a60db47/runs/c652a4bb-ea77-4ea6-9955-dfd877967b99/dcdfbc9f-b08f-49ab-9070-d206c5132c9d/1/logs?job=test

### Deploy regression test
Built a local `vlocity_build` from this branch and ran deploy + activate of 5 Claim
OmniScripts against a Fennia support sandbox — this exercises both changed paths at
runtime (`queryservice` during delta-check, and the `omniscript.js` activation
`searchKey` queries). Ran once for the queryservice fix and again after the
consolidation.

Both runs: clean success — 5/5 deployed, 0 errors, no `Too many SOQL` async failures,
all 5 scripts activated at their new version with a compilation row each, and no
runtime error resolving `escapeSOQLString`. Org returned to baseline afterward. (A
real deploy uses quote-free values, so injection-neutralization itself stays proven
by the unit tests above.)

### Notes
- Behavior is unchanged for legitimate values (no quotes/backslashes).
- No unrelated changes.


sf-pallavi-das (approved):
LGTM

snehaljha-sf (approved):
LGTM

PR #808, opened 2026-09-15
@W-24114276@ Prevent path traversal in DataPack expansion via unsanitized dataPackType (CWE-22)

amaindoliya-sf:
### What

Fixes **CWE-22 path traversal** in the classic (non-SFDX) DataPack **expansion** path (`W-24114276`).

`DataPacksExpand.generateFolderPath(dataPackType, parentName)` in `lib/datapacksexpand.js` joined `dataPackType` — which comes from untrusted DataPack content (`VlocityDataPackType`) and, unlike `parentName`/filenames, is **not** run through `generateFolderOrFilename` sanitization — directly into an on-disk path under `this.targetPath`. `path.normalize` collapses `../` but does not confine the result, so a crafted `VlocityDataPackType` (e.g. `../../../../tmp/evil`) escapes `targetPath`. Every classic-expansion filesystem sink routes through this one function (`generateFilepath` → `generateFolderPath`, plus the direct `ensureDir`/`emptyDir` call), so a crafted type enabled arbitrary file **write** (`fs.outputFile` in `writeFile`), arbitrary **read** (parent/relationship-key `fs.readFile`), and directory-content **deletion** (`fs.emptyDir`).

### Fix

- Add `assertPathWithin(baseDir, targetPath)` — resolves both paths and throws if the target is not `baseDir` itself or under `baseDir + path.sep`. Logs via `VlocityUtils.error` before throwing so blocked attempts leave an audit trail even at call sites that swallow expansion errors.
- Call it inside `generateFolderPath` — the single chokepoint for every classic-expansion read/write path — so a crafted `VlocityDataPackType` can no longer escape `targetPath`.

The sanitized filename tail is already safe (`generateFolderOrFilename` maps `.`→`-`, so `..` cannot survive; leading/trailing separators stripped) — the hole was exclusively the unsanitized `dataPackType` folder component.

Scope note: the sibling `expandSFDX` sink is addressed separately under **W-24123540**; this PR intentionally covers only the classic expansion path.

### Tests

`test/datapacksexpand.spec.js` — new `generateFolderPath (path-traversal guard)` suite: legit `dataPackType` stays inside `targetPath`; a `../…` escape throws; a sibling-directory prefix escape (`targetPath` + `-evil`) throws. Full `npm run unitTest` green.


amaindoliya-sf:
Superseded by #811, which combines this fix with W-24123540 behind the shared enforceSecurityValidations flag (default on) and targets alpha.

PR #810, opened 2026-09-15
@W-24098246: escape shell-interpolated paths to prevent OS command…

snehaljha-sf:
… injection

sfdx._orgLogin() (lib/sfdx.js) interpolated authFilePath unquoted into a shell.exec() command string, unlike every other parameter in the file which uses "${param}". Several execSync() call sites in utilityservice.js (getGitDiffsFromOrgToLocal) and datapacksjob.js interpolated jobInfo.projectPath/expansionPath and hashKey the same unquoted way. Any of these values containing shell metacharacters (;, $(), backticks, &&, |) executes arbitrary commands with the privileges of the build process (CWE-78) - e.g. a malicious project path or property value from a crafted PR.

Added sfdx.shellEscapeArg(), a single-quote shell-escaping helper (the standard POSIX technique: close quote, insert escaped quote, reopen quote), and applied it at every unescaped interpolation site of these values. Chose this over child_process.execFileSync() (the alternative the security scan suggested) because these callers build multi-segment shell command strings (cd X && git ...) rather than a single argv array, so switching to execFileSync would require restructuring each call site; escaping the untrusted segment is the minimal fix that closes the injection without changing command semantics.

Scope: this ticket's scan also flagged the same root cause in via_components and vloc_release (5 other chains) - out of scope here, tracked separately per repo.


sf-pallavi-das (approved):

snehaljha-sf:
Put changes behind the flag: enableShellArgEscaping

How to use the flag:
- Job YAML / options: enableShellArgEscaping: true (default false, set in lib/defaultjobsettings.yaml)
- CLI: vlocity packUpdateSettings ... --no-enableShellArgEscaping (or --enableShellArgEscaping=true)

PR #811, opened 2026-09-15
@W-24123540@ @W-24114276@ Prevent path traversal in DataPack expansion (CWE-22), gated by enforceSecurityValidations

amaindoliya-sf:
## Summary

Combined fix for two related path-traversal (CWE-22) vulnerabilities in DataPack expansion, both routed through a single shared guard and gated behind one **opt-in** flag (`enforceSecurityValidations`, **default OFF**).

### W-24123540 — arbitrary file write in `expandSFDX`
`expandSFDX` builds the write target from `dataPack.VlocityDataPackKey` (and, for static resources, `resourceFile.filePath`) taken directly from bundle JSON. A crafted bundle with a key like `../../../ESCAPED/pwned.json` caused `path.join` to resolve **outside** the SFDX project directory, writing attacker-controlled content anywhere the process could write.

### W-24114276 — path traversal via unsanitized `dataPackType`
`generateFolderPath` joins the untrusted `VlocityDataPackType` into `targetPath` without sanitization, so a crafted type (`../../…`) could escape the expansion target directory.

## Fix

- Added a shared `DataPacksExpand.prototype.assertPathWithin(baseDir, targetPath)` helper: lexically resolves both paths and throws unless the target stays under the base (`=== base` or `startsWith(base + path.sep)`). It logs before throwing because some callers swallow expansion errors.
- Enforced at all three sinks: `generateFolderPath`, the `expandSFDX` file write, and the `expandSFDX` static-resource copy.

## Opt-in flag (`enforceSecurityValidations`, default OFF)

Per the current rollout plan, the guard ships **disabled by default** so existing customers see no behavior change, and can opt in to enforcement — first in a **sandbox** — before it becomes the default in a later release. This gives customers time to confirm no legitimate workflow of theirs relies on writing outside the project directory.

- Default `false` in `lib/defaultjobsettings.yaml` (merged via the existing `mergeInJobDefaults` precedent — the default only applies when the key is `undefined`, so an explicit value always wins).
- Read as `this.enforceSecurityValidations === true` inside `assertPathWithin`, set once from `jobInfo` in `expand()`. The guard fires **only** when the flag is explicitly `true`; absent or `false` → not enforced.
- Enable via `-enforceSecurityValidations true` on the CLI (added to `VLOCITY_COMMANDLINE_OPTIONS`), or `enforceSecurityValidations: true` in the job YAML.
- The flag is operator-controlled (job YAML / CLI), **not** reachable from bundle content — an attacker importing a malicious bundle cannot toggle the guard.

### Unit tests
`npm run unitTest` → **65 passing**. Includes the path-traversal guard specs for both WIs (run with enforcement enabled) plus an `enforceSecurityValidations opt-in` block asserting: the guard is **not** enforced when the flag is `undefined` (default off) or explicitly `false`, and is enforced **only** when explicitly `true`.

### Live end-to-end through the real `packExpandFile` command

The guard's enforcement path was validated live against the real CLI by expanding a malicious DataPack whose `VlocityDataPackKey` was `../../../../ESCAPED/pwned.json` — a key that resolves to a location **outside** the SFDX project folder, so the out-of-project file appearing means the traversal succeeded. The observed behavior maps directly onto the flag state:

- **Flag enabled (`enforceSecurityValidations: true` via job YAML, or `-enforceSecurityValidations true` on the CLI):** the tool logged `Path Traversal Blocked` / `Path traversal detected: refusing to write outside target directory` and **wrote nothing** outside the project. CLI override takes precedence over the job YAML.
- **Flag absent (default) or explicitly `false`:** the guard did not fire; the tool logged `Writing SFDX File …/ESCAPED/pwned.json` and the file **was written** outside the project.

**Net:** with the default-off flag, out-of-the-box behavior is unchanged from today (no enforcement); a customer opts in via YAML or CLI — ideally in a sandbox first — to turn the guard on. We also verified `-enforceSecurityValidations true` parses to boolean `true` and is copied into `jobInfo` by the existing CLI override loop, while an absent flag falls back to the `false` default.

### CloudBees Test
* Success: https://cloudbees.io/vlocityinc/components/670f856a-778d-4d66-9417-519d4a60db47/runs/19c39fdc-833b-4bb5-8197-bc90c469e810/4986c74d-4a42-468b-8a81-5f187fcf9bf4/1/logs?componentId=670f856a-778d-4d66-9417-519d4a60db47&workflowId=19c39fdc-833b-4bb5-8197-bc90c469e810&job=test&step=s003-1

> Note: CloudBees link above was from the prior (default-on) revision; re-run against this revision before merge.


sf-pallavi-das (approved):
LGTM

lib/defaultjobsettings.yaml
```diff
@@ -31,6 +31,7 @@ supportHeadersOnly: true
 allowHeadersPartial: true
 supportForceDeploy: true
 strict: false
+enforceSecurityValidations: true
```
snehaljha-sf, line 34: I see here the flag named enforce (with true it's enforced), but in code we're checking just for false (missing flag would mean true) this is something unconventional if we compare it with other flags we have in place. We can negate it to something like RevertSecurityValidation or SkipSecurityValidation

snehaljha-sf (approved):
One minor comment (which can cause confusion). Rest LGTM

lib/defaultjobsettings.yaml
amaindoliya-sf: The code has both conventions. In the same file we have `(jobInfo.useAllRelationships !== false && …)` convention. Since this is a transient change, and eventually we are going to delete the flag, I am keeping it as is.

